PR #15, opened 2024-02-06
run docker jobs without network and check for CVE 2022 0492 before ma…

mpnowacki-reef:
…rking executor as ready

### DIFF
--- a/executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
+++ b/executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
@@ -160,17 +160,26 @@ class JobRunner:
             raise JobError(msg)
 
     async def run_job(self, job_request: V0JobRequest):
+        msg = None
         for arg in job_request.docker_run_options:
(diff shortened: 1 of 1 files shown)


executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -247,12 +274,41 @@ def __init__(self, *a, **kw):
...
+
+        return True
+
     async def _executor_loop(self):
         logger.debug(f'Connecting to miner: {settings.MINER_ADDRESS}')
         async with self.miner_client:
             logger.debug(f'Connected to miner: {settings.MINER_ADDRESS}')
-
             initial_message: V0InitialJobRequest = await self.miner_client.initial_msg
+            logger.debug('Checking for CVE-2022-0492 vulnerability')
```
ppolewicz, line 308: I think this should go into the checker function, otherwise to see this consistently all users would have to copy this line before the checker

PR #30, opened 2024-02-25
Manage dependencies with pdm

vbaltrusaitis-reef:
This PR pulls the cruft template updates which introduce pdm fro dependency management. This mostly scripts, noxfiles, CI, and Dockerfiles. The most important changes:

- Python dependencies are now installed with lockfiles (pdm.lock), guaranteeing that exactly the same package versions will be installed across distributions.
- Miner, executor, validator Docker image is not build with `./miner`, `./executor`, `./validator` as roots, not `./miner/app` etc. This is so that Docker build context sees `pyproject.toml`.
- All dependencies (including development) are now in `pyproject.toml` files, not across `noxfile.py` scripts.


vbaltrusaitis-reef:
Integration test fails because GitHub runner runs out of SSD (the default runner has 16GB or space or so) when installing dependencies for miner, validator, executor separately. It might be possible to make the process save less space but I think it might be better to use a bigger runner for that particular test.

.github/workflows/executor_ci.yml
```diff
@@ -45,9 +45,12 @@ jobs:
...
           cache: "pip"
+      - name: Install dependencies
+        run: python -m pip install --upgrade nox pdm
       - name: Prepare environment
         run: ./setup-dev.sh
-      - name: Install dependencies
-        run: python -m pip install --upgrade nox pip setuptools
+      - name: Build backenddevelopersltd/compute-horde-job-echo:v0-latest image
+        run: docker build . -t backenddevelopersltd/compute-horde-job-echo:v0-latest
+        working-directory: ./executor/app/src/compute_horde_executor/executor/tests/integration/docker_image_for_tests
```
vbaltrusaitis-reef, line 54: Image `backenddevelopersltd/compute-horde-job-echo:v0-latest` in the docker hub currently targets arm64 and doesn't work on GH action runners. So for the time being I'm building the image in the actions.

executor/app/build-image.sh
```diff
@@ -2,4 +2,5 @@
 
 IMAGE_NAME="backenddevelopersltd/compute-horde-executor:v0-latest"
 rsync -avzP ../../compute_horde packages/
-docker build -t $IMAGE_NAME -f envs/prod/Dockerfile .
+# Dockerfile assumes that build is run in the parent dir
+cd .. && docker build -t $IMAGE_NAME -f app/envs/prod/Dockerfile .
```
agoncharov-reef, line 6: Thought: I would try to avoid sending a lot of context just because of one `pdm.lock` file. I would create `pdm.lock` inside a docker context folder, and make a symlink in root folder.

executor/app/envs/prod/Dockerfile
```diff
@@ -5,20 +5,26 @@ FROM $BASE_IMAGE AS base-image
...
-COPY ./envs/prod/gunicorn.conf.py /root/src/
-COPY ./envs/prod/celery-entrypoint.sh /root/src/
+COPY pyproject.toml pdm.lock ./
+RUN pdm lock --check
+COPY ./app/packages/compute_horde /root/compute_horde
+
+RUN --mount=type=cache,target=/tmp/pdm_cache \
+  pdm config cache_dir /tmp/pdm_cache && \
+  pdm config python.use_venv False && \
+  pdm sync --prod --group :all
```
agoncharov-reef, line 19: Not related to your changes, but I think `app/packages/compute_horde` will change more often than requirements, so these `COPY` and `RUN` lines should swap.

Well, and `ENV` lines below, too. They are unlikely to change ever, why not putting them closer to the beginning of `Dockerfile`?

agoncharov-reef (approved):

executor/app/build-image.sh
vbaltrusaitis-reef: It's `pyproject.toml` as well (not that it matters for you argument, really). The reason I decided against creating a symlink or copying the file in a build step or any of those things is that `pyproject.toml` is not a self-contained piece of information (like requirements.txt), it has meaning that depends on where it's positioned relative to other files (for example configuration sections for tools like pytest might contain relative paths there). Copying or having symlinks to me just seems more confusing, it's not clear where the project root is just by looking at the file tree.

If anything, I'd rather make the structure flatter and extract docker-related files to a separate `/docker` folder.

I suspect this won't entirely convince you but is this a blocker? This change (building from the root) was already approved and merged in cookiecutter template. I do agree that this is annoying though.

executor/app/envs/prod/Dockerfile
vbaltrusaitis-reef: In the previous iteration of Dockerfile, `compute_horde` was copied before installing from `requirements.txt` because `requirements.txt` actually contained `../../../compute_horde` - this package is was installed via pip, swapping would not be enough. The same is the case with pdm now, even more so. I guess it would be possible and perhaps desirable to first copy only `../compute_horde/pyproject.toml`, install everything with `pdm sync` and _then_ copy and again install the real compute_horde package. But this felt like a new optimization, I didn't want to include it into this PR.

.github/workflows/executor_ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           python-version: ${{ env.PYTHON_DEFAULT_VERSION }}
           cache: "pip"
       - name: Install dependencies
-        run: python -m pip install --upgrade nox pip setuptools
+        run: python -m pip install --upgrade nox pdm
```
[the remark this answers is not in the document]
vbaltrusaitis-reef: Fair enough, will change this. I was going back-and-forth between limiting the version and not. The argument for not fixing the version was that breaking changes only matter if they actually break the workflows here, which would be immediately noticed, unlike some breaking change in an actual Python dependency of the package. But I'm already fixing pdm version in `Dockerfile`, I should at least be consistent.

executor/app/envs/prod/Dockerfile
agoncharov-reef: Ah yes you're right, sorry

executor/app/build-image.sh
agoncharov-reef: No, it's just a thought. Not a blocker. Thanks for explanation :)

PR #31, opened 2024-02-26
Small miner README improvements

przadka:


miner/envs/runner/README.md
```diff
@@ -58,6 +56,8 @@ services:
 
 ```
 
+With the `.env` and `docker-compose.yml` already in place, you can start the miner with `docker-compose up -d`. Test if it is running properly with `docker-compose ps`.
```
mjurbanski-reef, line 59: sembr

przadka: Fixed this in two places

PR #41, opened 2024-03-12
Staging runner of validator/miner/executor and other fixes

emnoor-reef:
Copied `validator/envs/runner` to `validator/envs/runner-staging` and adjusted.
Copied `miner/envs/runner` to `miner/envs/runner-staging` and adjusted.
I built the staging images locally by adding `-staging` to the `IMAGE_NAME` variables of `*/app/build-image.sh` files.


executor/app/src/compute_horde_executor/settings.py
```diff
@@ -249,8 +249,8 @@ def wrapped(*args, **kwargs):
 
 MINER_ADDRESS = env.str('MINER_ADDRESS')
 EXECUTOR_TOKEN = env.str('EXECUTOR_TOKEN')
-VOLUME_MAX_SIZE_BYTES = env.int('VOLUME_MAX_SIZE_BYTES')
-OUTPUT_ZIP_UPLOAD_MAX_SIZE_BYTES = env.int('OUTPUT_ZIP_UPLOAD_MAX_SIZE_BYTES')
+VOLUME_MAX_SIZE_BYTES = env.int('VOLUME_MAX_SIZE_BYTES', default=104857600)  # 100MB
+OUTPUT_ZIP_UPLOAD_MAX_SIZE_BYTES = env.int('OUTPUT_ZIP_UPLOAD_MAX_SIZE_BYTES', default=314572800)  # 300MB
```
agoncharov-reef, line 253: https://github.com/oerd/reef-handbook/blob/master/docs/Code_conventions.md#dont-use-stuff-from-forbidden-technology-list-without-prior-discussion

miner/app/src/compute_horde_miner/miner/executor_manager/docker.py
```diff
@@ -41,6 +40,6 @@ async def reserve_executor(self, token):
             # the executor must be able to spawn images on host
             "-v", "/var/run/docker.sock:/var/run/docker.sock",
             "-v", "/tmp:/tmp",
-            EXECUTOR_IMAGE,
+            settings.EXECUTOR_IMAGE,
```
agoncharov-reef, line 43: Correct me if I'm wrong.
Previously our code was stitched to executor image version:
```
commit 123 <---> image-v0-latest
commit 134 <---> image-v1-latest
```
So in order to switch to new version we had to just update version number in code and deploy. With the change introduced, we have to not forget to change environment variable and restart/redeploy.

I don't quite understand the benefit `¯\_(ツ)_/¯`

agoncharov-reef (commented):
I don't understand why copying entire runner folder for staging environment instead of making places that differ depend on env vars and then just switch symlinks from `.env -> .env-prod` to `.env -> .env-staging`?

miner/app/src/compute_horde_miner/miner/executor_manager/docker.py
emnoor-reef: Then how do you differentiate between prod executor and staging executor?

executor/app/src/compute_horde_executor/settings.py
emnoor-reef: https://reeftechnologies.slack.com/archives/C06DMMT987M/p1706390531949159

The executor is being spawned dynamically in docker would mean we need to list the envs here, which is worse IMO: https://github.com/backend-developers-ltd/ComputeHorde/blob/staging/miner/app/src/compute_horde_miner/miner/executor_manager/docker.py#L36-L45

emnoor-reef:
I got lazy on it and copied the whole folder. I am merging this, as other fixes needs the changes of this branch. I will fix the copy-pasta staging runner later.

PR #42, opened 2024-03-13
Make job time wait limit shared between 'preparation' and 'running' stages

agoncharov-reef:
Ref: https://reeftechnologies.slack.com/archives/C06DMMT987M/p1710323555348469?thread_ts=1710288682.424139&cid=C06DMMT987M

NOT TESTED!

### DIFF
--- a/validator/app/src/compute_horde_validator/validator/utils.py
+++ b/validator/app/src/compute_horde_validator/validator/utils.py
@@ -1,0 +1,19 @@
+import time
+from collections.abc import Callable
+from datetime import timedelta
+
+
+class Timer:
+    def __init__(self, name: str = "", output_fn: Callable | None = None):
+        self.name = name
+        self.output_fn = output_fn
+
+    def __enter__(self):
+        self.start = time.time()
+        return self
+
+    def __exit__(self, *args):
+        self.end = time.time()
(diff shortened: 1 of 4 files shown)


validator/app/src/compute_horde_validator/validator/utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def __enter__(self):
+        self.start = time.time()
+        return self
+
+    def __exit__(self, *args):
+        self.end = time.time()
+        self.elapsed = timedelta(seconds=self.end - self.start)
+        if self.output_fn:
+            self.output_fn(f"{self.name} took {self.interval:.2f} seconds")
```
mjurbanski-reef, line 19:
```suggestion
            self.output_fn(f"{self.name} took {self.elapsed:.2f} seconds")
```

but IMO output_fn used this way doesn't seem versatile enough to be really useful - for example for logging you would like either structured log or at least %formatting 

maybe if it was just:
```suggestion
            self.output_fn(self)
```
then you could do whatever you wanted

agoncharov-reef: download

mzukowski-reef (changes requested):
please don't merge - I want a little bit different logic - I will introduce it myself with miner executor manager queue change

PR #46, opened 2024-03-15
Check download size while streaming download if Content-Length is missing

emnoor-reef:


executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -389,11 +389,16 @@ async def _unpack_volume(self, job_request: V0JobRequest):
             with tempfile.NamedTemporaryFile() as download_file:
                 async with httpx.AsyncClient() as client:
                     async with client.stream('GET', job_request.volume.contents) as response:
-                        volume_size = int(response.headers["Content-Length"])
-                        if 0 < settings.VOLUME_MAX_SIZE_BYTES < volume_size:
-                            raise JobError("Input volume too large")
+                        if (content_length := response.headers.get("Content-Length")) is not None:
+                            # check size early if Content-Length is present
+                            if 0 < settings.VOLUME_MAX_SIZE_BYTES < int(content_length):
```
mjurbanski-reef, line 394:
```suggestion
                            if settings.VOLUME_MAX_SIZE_BYTES < int(content_length):
```
and same below

executor/app/src/compute_horde_executor/executor/tests/integration/test_main_loop.py
```diff
@@ -187,6 +187,110 @@ def test_zip_url_too_big_volume_should_fail(httpx_mock: HTTPXMock, settings):
     ]
 
 
+def test_zip_url_volume_without_content_length(httpx_mock: HTTPXMock):
+    zip_url = 'https://localhost/payload.txt'
+
+    def response_callback(request):
+        import httpx
```
mjurbanski-reef, line 194: nitpick: inline import

executor/app/src/compute_horde_executor/executor/tests/integration/test_main_loop.py
```diff
@@ -187,6 +187,110 @@ def test_zip_url_too_big_volume_should_fail(httpx_mock: HTTPXMock, settings):
...
+        response = httpx.Response(
+            status_code=200,
+            extensions={"http_version": "HTTP/1.1".encode("ascii")},
```
mjurbanski-reef, line 197:
```suggestion
            extensions={"http_version": b"HTTP/1.1"},
```

executor/app/src/compute_horde_executor/executor/tests/integration/test_main_loop.py
```diff
@@ -187,6 +187,110 @@ def test_zip_url_too_big_volume_should_fail(httpx_mock: HTTPXMock, settings):
...
+        {
+            "message_type": "V0ReadyRequest",
+            "job_uuid": job_uuid,
+        },
+        {
+            "message_type": "V0FailedRequest",
+            "docker_process_exit_status": None,
+            "timeout": False,
+            "docker_process_stdout": "Input volume too large",
+            "docker_process_stderr": "",
```
mjurbanski-reef, line 288: Im surprised to see errors printed to stdout instead of stderr but I don't know how it is supposed to work

it kinda seems like it wasn't docker process output anyhow, which makes it doubly strange to me

mjurbanski-reef (approved):

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
emnoor-reef: This is here to make size checks optional. Setting `settings.VOLUME_MAX_SIZE_BYTES <= 0` would disable these checks.

executor/app/src/compute_horde_executor/executor/tests/integration/test_main_loop.py
emnoor-reef: For some weird reasons that's how it has been. Errors occurred before starting the docker process is sent to `docker_process_stdout` field. :|

PR #59, opened 2024-03-27
Docker image LRU eviction in miner (executor)

emnoor-reef:
- Remove the logic of docker image clean-up after job execution.
- `docuum` keeps track of docker events. When docker image storage crosses a threshold it evicts the least recently used images.
- Image storage threshold can be configured with the `DOCKER_STORAGE_THRESHOLD` environment variable.
- A default threshold of `50 GB` is arbitrarily chosen.

### DIFF
--- a/miner/envs/runner-staging/data/docker-compose.yml
+++ b/miner/envs/runner-staging/data/docker-compose.yml
@@ -123,7 +123,6 @@ services:
       "--keep", "^backenddevelopersltd/compute-horde-job:v0-latest",
       "--keep", "^us-central1-docker.pkg.dev/twistlock-secresearch/public/can-ctr-escape-cve-2022-0492:latest",
     ]
-    pull_policy: always
     init: true
     restart: unless-stopped
     env_file: ./.env
--- a/miner/envs/runner-staging/data/docker-compose.yml
+++ b/miner/envs/runner-staging/data/docker-compose.yml
@@ -115,9 +115,12 @@ services:
       "--threshold", "${DOCKER_STORAGE_THRESHOLD:-50 GB}",
       "--keep", "^redis:6-alpine",
       "--keep", "^postgres:14.0-alpine",
+      "--keep", "^containrrr/watchtower",
+      "--keep", "^stephanmisc/docuum",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-runner-staging:v0-latest",
       "--keep", "^backenddevelopersltd/compute-horde-miner-staging:v0-latest",
       "--keep", "^backenddevelopersltd/compute-horde-miner-nginx:v0-latest",
-      "--keep", "^backenddevelopersltd/compute-horde-job",
+      "--keep", "^backenddevelopersltd/compute-horde-job:v0-latest",
       "--keep", "^us-central1-docker.pkg.dev/twistlock-secresearch/public/can-ctr-escape-cve-2022-0492:latest",
     ]
     pull_policy: always
--- a/miner/envs/runner-staging/data/docker-compose.yml
+++ b/miner/envs/runner-staging/data/docker-compose.yml
@@ -111,7 +111,14 @@ services:
 
   docuum:
     image: stephanmisc/docuum
-    command: --threshold ${DOCKER_STORAGE_THRESHOLD:-50 GB}
+    command: [
+      "--threshold", "${DOCKER_STORAGE_THRESHOLD:-50 GB}",
+      "--keep", "^redis:6-alpine",
+      "--keep", "^postgres:14.0-alpine",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-staging:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-nginx:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-job",
+    ]
     pull_policy: always
     init: true
     restart: unless-stopped
--- a/miner/envs/runner-staging/data/docker-compose.yml
+++ b/miner/envs/runner-staging/data/docker-compose.yml
@@ -110,13 +110,13 @@ services:
       - "com.centurylinklabs.watchtower.enable=true"
 
   docuum:
-    image: stephanmisc/docuum
+    image: stephanmisc/docuum:0.23.1
     command: [
       "--threshold", "${DOCKER_STORAGE_THRESHOLD:-50 GB}",
       "--keep", "^redis:6-alpine",
       "--keep", "^postgres:14.0-alpine",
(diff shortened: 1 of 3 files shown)


miner/envs/runner-staging/data/docker-compose.yml
```diff
@@ -109,7 +109,34 @@ services:
...
+      "--threshold", "${DOCKER_STORAGE_THRESHOLD:-50 GB}",
+      "--keep", "^redis:6-alpine",
+      "--keep", "^postgres:14.0-alpine",
+      "--keep", "^containrrr/watchtower",
+      "--keep", "^stephanmisc/docuum",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-runner-staging:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-staging:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-nginx:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-job:v0-latest",
+      "--keep", "^us-central1-docker.pkg.dev/twistlock-secresearch/public/can-ctr-escape-cve-2022-0492:latest",
```
mjurbanski-reef, line 124: why are these necessary? aren't they in constant use therefor docuum should never evict them?

miner/envs/runner/data/docker-compose.yml
```diff
@@ -109,7 +109,34 @@ services:
...
+      "--threshold", "${DOCKER_STORAGE_THRESHOLD:-50 GB}",
+      "--keep", "^redis:6-alpine",
+      "--keep", "^postgres:14.0-alpine",
+      "--keep", "^containrrr/watchtower",
+      "--keep", "^stephanmisc/docuum",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-runner:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-miner:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-nginx:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-job:v0-latest",
+      "--keep", "^us-central1-docker.pkg.dev/twistlock-secresearch/public/can-ctr-escape-cve-2022-0492:latest",
```
mjurbanski-reef, line 124: ditto

miner/envs/runner-staging/data/docker-compose.yml
```diff
@@ -109,7 +109,34 @@ services:
...
+    init: true
+    restart: unless-stopped
+    env_file: ./.env
+    volumes:
+      - docuum:/root
+      - /var/run/docker.sock:/var/run/docker.sock
+    logging:
+      <<: *logging
+    labels:
+      - "com.centurylinklabs.watchtower.enable=true"
```
mjurbanski-reef, line 136: As with any external service (db, redis) please remove autoupdate and specify strict image version to be used.

We do not want this to become broken with an automatic update and there are security considerations as well. Perhaps most importantly - there are not benefits to autoupdate it.

miner/envs/runner/data/docker-compose.yml
```diff
@@ -109,7 +109,34 @@ services:
...
+    init: true
+    restart: unless-stopped
+    env_file: ./.env
+    volumes:
+      - docuum:/root
+      - /var/run/docker.sock:/var/run/docker.sock
+    logging:
+      <<: *logging
+    labels:
+      - "com.centurylinklabs.watchtower.enable=true"
```
mjurbanski-reef, line 136: ditto

mjurbanski-reef (changes requested):

miner/envs/runner-staging/data/docker-compose.yml
emnoor-reef: In my testing, `docuum` tries to delete these and fails, over and over.

I guess it happens because these containers are started before `docuum`, so it has no idea about them. And maybe it doesn't look at currently running containers.

mjurbanski-reef: that is an annoying flaw in docuum, I guess we could just ignore errors since they should be resolved after next restart, but your solution works as well

miner/envs/runner-staging/data/docker-compose.yml
```diff
@@ -109,7 +109,32 @@ services:
...
+    command: [
+      "--threshold", "${DOCKER_STORAGE_THRESHOLD:-50 GB}",
+      "--keep", "^redis:6-alpine",
+      "--keep", "^postgres:14.0-alpine",
+      "--keep", "^containrrr/watchtower",
+      "--keep", "^stephanmisc/docuum:0.23.1",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-runner-staging:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-staging:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-miner-nginx:v0-latest",
+      "--keep", "^backenddevelopersltd/compute-horde-job:v0-latest",
```
mjurbanski-reef, line 123:
```suggestion
      "--keep", "^backenddevelopersltd/compute-horde-.*:v0-latest",
```

miner/envs/runner-staging/data/docker-compose.yml
```diff
@@ -109,7 +109,32 @@ services:
...
+      "--keep", "^us-central1-docker.pkg.dev/twistlock-secresearch/public/can-ctr-escape-cve-2022-0492:latest",
+    ]
+    pull_policy: always
```
mjurbanski-reef, line 126: please remove pull_policy from here

miner/envs/runner-staging/data/docker-compose.yml
```diff
@@ -109,7 +109,32 @@ services:
...
+    init: true
+    restart: unless-stopped
+    env_file: ./.env
```
mjurbanski-reef, line 129: question(security): do we need to share all our envs with it?

mjurbanski-reef (approved):

miner/envs/runner-staging/data/docker-compose.yml
emnoor-reef: That was a oversight from my part. We don't want to share any environment variables with it. The only var related to this, `DOCKER_STORAGE_THRESHOLD`, is already used in the `command`.

miner/envs/runner-staging/data/docker-compose.yml
emnoor-reef: We might want create on-off jobs. So wildcard marking them as "keep" could be problematic. For now, let's keep them specified. We can always trim them later with a wildcard if we see this list growing.

PR #61, opened 2024-04-01
Handle miner connection refusal in validator

emnoor-reef:
The validators/miners trying to connect a miner that is refusing the connection can lead to an infinite loop, because miner connection was retried forever (with exponential backoff). As a consequence, the status of the job is never reported in the facilitator.

This PR adds a max-retry limit of 5 for connecting to a miner. If the limit is exceeded, the job is considered failed and reported as such to the facilitator.

Previously on facilitator:
image

After this PR:
image

**Note:** the `MinerConnectionError` exception is purposefully not handled in the executor, so that miners can notice fast if their executors cannot connect to their miners. This is often an issue with miner's configuration.

### DIFF
--- a/compute_horde/compute_horde/miner_client/base.py
+++ b/compute_horde/compute_horde/miner_client/base.py
@@ -79,11 +79,14 @@ class AbstractMinerClient(abc.ABC):
                     sleep_time = self.sleep_time()
                     logger.info(f'Retrying connection to miner {self.miner_name} in {sleep_time:0.2f}')
                     await asyncio.sleep(sleep_time)
-                self.debounce_counter += 1
                 self.ws = await self._connect()
                 self.read_messages_task = self.loop.create_task(self.read_messages())
+                if self.debounce_counter:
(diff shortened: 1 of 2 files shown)


mjurbanski-reef (commented):
it does seem like improvement, but I would like to note there seem to be similar problem in `send_model`

at which point, debounce_counter would also be useful to have. That also means - that in the code that you are currently modifying that debounce_counter should be reset upon successful connection (perhaps log the fact of success + debounce_counter value pre reset?)

emnoor-reef:
On shit. All the places `send_model` is used, is under the assumption that it will not fail. Need to handle it in a lotta places. 😵‍💫

mzukowski-reef:
> On shit. All the places `send_model` is used, is under the assumption that it will not fail. Need to handle it in a lotta places. 😵‍💫

I thought that "Create persistent message queue for executor<>miner and miner<>validator communication" task would handle this part...

PR #68, opened 2024-04-12
Admin panel

andreea-popescu-reef:
https://reeftechnologies.atlassian.net/browse/COM-67
https://reeftechnologies.atlassian.net/browse/COM-51
https://reeftechnologies.atlassian.net/browse/COM-50
https://reeftechnologies.atlassian.net/browse/COM-65


miner/app/src/compute_horde_miner/miner/admin.py
```diff
@@ -1,7 +1,48 @@
...
+admin.site.register(AcceptedJob, admin_class=ReadOnlyAdmin)
+admin.site.register(Validator, admin_class=ReadOnlyAdmin)
+admin.site.register(ValidatorBlacklist)
+
+def maybe_create_default_admin():
+    # Create default admin user if missing
+    try:
+        admin_user_exists = User.objects.filter(is_superuser=True).exists()
+    except Exception:
+        # If database not setup skip creating default admin
```
agoncharov-reef, line 37: Is it possible that DB is not set up?
If yes then it's better to catch specific exception, not `Exception`.

miner/app/src/compute_horde_miner/miner/admin.py
```diff
@@ -1,7 +1,48 @@
...
+        return
+    if not admin_user_exists:
+        admin_password = os.getenv("DEFAULT_ADMIN_PASSWORD")
```
agoncharov-reef, line 40: Nitpick: I'd define it in settings (`DEFAULT_ADMIN_PASSWORD = env("DEFAULT_ADMIN_PASSWORD", default=None)`), so that here we use `settings.DEFAULT_ADMIN_PASSWORD`. I think defining settings in `settings.py` is better than reading settings from env vars in random place of an application.

miner/app/src/compute_horde_miner/miner/admin.py
```diff
@@ -1,7 +1,48 @@
...
+        if admin_password is None:
+            logger.warning("Not creating Admin user - please set DEFAULT_ADMIN_PASSWORD env variable")
+        else:
+            logger.info("Creating Admin user with DEFAULT_ADMIN_PASSWORD")
+            admin_user = User.objects.create_superuser(username='admin', email='admin@admin.com', password=admin_password)
+            admin_user.save()
```
agoncharov-reef, line 46: It's redundant:
> create_superuser(username, email=None, password=None, **extra_fields)[¶](https://docs.djangoproject.com/en/5.0/ref/contrib/auth/#django.contrib.auth.models.UserManager.create_superuser)
> Creates, saves and returns a [User](https://docs.djangoproject.com/en/5.0/ref/contrib/auth/#django.contrib.auth.models.User).

miner/app/src/compute_horde_miner/miner/admin.py
```diff
@@ -1,7 +1,48 @@
...
+
+maybe_create_default_admin()
```
agoncharov-reef, line 48: I think all this should go to `app.ready()`, something like in this question description: https://stackoverflow.com/questions/60562025/automatically-create-django-model-instances-at-startup-with-empty-database

miner/app/src/compute_horde_miner/miner/miner_consumer/validator_interface.py
```diff
@@ -138,6 +138,12 @@ async def handle(self, msg: BaseValidatorRequest):
             self.msg_queue.append(msg)
             return
         if isinstance(msg, validator_requests.V0InitialJobRequest):
+            validator_blacklisted = await ValidatorBlacklist.objects.filter(validator=self.validator).aexists()
+            if validator_blacklisted:
+                logger.info(f"Declining job {msg.job_uuid} from blacklisted validator: self.validator_key")
```
agoncharov-reef, line 143: f-string missing interpolation near `self.validator_key` :P

miner/app/src/compute_horde_miner/miner/models.py
```diff
@@ -12,11 +12,18 @@ def default(self, obj):
...
     active = models.BooleanField()
 
+    def __str__(self):
+        return self.public_key
+
+class ValidatorBlacklist(models.Model):
+    validator = models.OneToOneField(Validator, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return self.validator.public_key
```
agoncharov-reef, line 26: Nitpick: Since the model has only one field, it could be just model field (`is_blacklisted`)

validator/app/src/compute_horde_validator/validator/admin.py
```diff
@@ -1,7 +1,49 @@
...
+    except Exception:
+        # If database not setup skip creating default admin
+        return
+    if not admin_user_exists:
+        admin_password = os.getenv("DEFAULT_ADMIN_PASSWORD")
+        if admin_password is None:
+            logger.warning("Not creating Admin user - please set DEFAULT_ADMIN_PASSWORD env variable")
+        else:
+            logger.info("Creating Admin user with DEFAULT_ADMIN_PASSWORD")
+            admin_user = User.objects.create_superuser(username='admin', email='admin@admin.com', password=admin_password)
```
agoncharov-reef, line 46: Better put those values in `settings.py`, magic strings aren't good

validator/app/src/compute_horde_validator/validator/admin.py
```diff
@@ -1,7 +1,49 @@
...
+            admin_user.save()
+
+maybe_create_default_admin()
```
agoncharov-reef, line 49: + same notes as in miner's admin, won't copy here

validator/nginx/templates/default.conf.template
```diff
@@ -6,21 +6,21 @@
 #
 
 server {
-    listen 80 default_server;
+    listen 127.0.0.1:80 default_server;
```
agoncharov-reef, line 9: Maybe I misunderstand. So this gonna work only inside the container, right? How are users supposed to get there?

agoncharov-reef (commented):
Nice! :+1:

miner/app/src/compute_horde_miner/miner/models.py
agoncharov-reef: (no need for action here)

miner/app/src/compute_horde_miner/miner/admin.py
andreea-popescu-reef: it should no longer be possible after moving this to post migrate

validator/nginx/templates/default.conf.template
andreea-popescu-reef: ah no, that should have been on the docker-compose port - the admin panel should only be visible via ssh on the machine

validator/app/src/compute_horde_validator/validator/admin.py
```diff
@@ -1,7 +1,33 @@
...
+from django.contrib.auth.models import User  # noqa
+
+from compute_horde_validator.validator.models import Miner, OrganicJob, SyntheticJob
 
 
 admin.site.site_header = "compute_horde_validator Administration"
 admin.site.site_title = "compute_horde_validator"
 admin.site.index_title = "Welcome to compute_horde_validator Administration"
+
+class AddOnlyAdmin(admin.ModelAdmin):
```
mzukowski-reef, line 12: I know that this was specified in a task, but allowing creating jobs is wrong right now - there is no logic that would make it usable - we should not have "fake" features

mzukowski-reef (changes requested):

validator/app/src/compute_horde_validator/validator/admin.py
andreea-popescu-reef: I'll comment out the Jobs  views for now

mzukowski-reef (approved):

PR #70, opened 2024-04-17
Add installer scripts for miner & validator

emnoor-reef:
heredocs inside heredocs! (╯°□°）╯︵ ┻━┻

The scripts have a lot of repeated code. But I decided to keep them like that, as de-duplicating them would mean complicated templating/code-generation.
But we *may* want to do it anyway, if we *need* similar installer scripts for staging and preprod.

There's one `shellcheck` warning in the scripts:
```
❯ shellcheck install_validator.sh

In install_validator.sh line 25:
ssh "$SSH_DESTINATION" <<ENDSSH
                         ^----^ SC2087 (warning): Quote 'ENDSSH' to make here document expansions happen on the server side rather than on the client.

For more information:
  https://www.shellcheck.net/wiki/SC2087 -- Quote 'ENDSSH' to make here docum...

```
but that is intended. It could be circumvented by preparing the `tmpvars` file locally (in `/tmp` or something) and then transferring it. But I didn't bother.

BTW, `shellcheck` does seem to understand the heredoc is shell script going to remote (ssh). But it does not seem to lint the heredoc itself as shell code. ¯\_(ツ)_/¯

Also, I didn't know where to put them, so I put them in root. LMK where is a better place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ for miner in miners:
 
 # Running
 
-To quickly start a validator or miner, create an ubuntu server and execute the following command from your local machine (where you have your wallet files).
+To quickly start a validator or miner, create an Ubuntu Server and execute the following command from your local machine (where you have your wallet files).
 
 **Validator:**
 ```shell
--- a/README.md
+++ b/README.md
@@ -30,5 +30,20 @@ In February 2024 this will change - subnet will define more resource types andVa
 
 # Running
 
-Refer to [Miner runner README](miner/envs/runner/README.md) and to 
-[Validator runner README](validator/envs/runner/README.md)
+To quickly start a validator or miner, create an ubuntu server and execute the following command from your local machine (where you have your wallet files).
+
+**Validator:**
+```shell
+curl -sSfL https://github.com/backend-developers-ltd/ComputeHorde/raw/master/install_validator.sh | bash -s - SSH_DESTINATION HOTKEY_PATH
+```
+
+**Miner:**
+```shell
+curl -sSfL https://github.com/backend-developers-ltd/ComputeHorde/raw/master/install_miner.sh | bash -s - SSH_DESTINATION HOTKEY_PATH
+```
+
+Replace `SSH_DESTINATION` with your server's connection info (i.e. `username@1.2.3.4`)
+and `HOTKEY_PATH` with the path of your hotkey (i.e. `~/.bittensor/wallets/my-wallet/hotkeys/my-hotkey`).
+This script installs necessary tools in the server, copies the keys and starts the validator/miner with the corresponding runner and default config.
+
+If you want to change the default config, see [Validator runner README](validator/envs/runner/README.md) and [Miner runner README](miner/envs/runner/README.md) for details.
--- a/install_miner.sh
+++ b/install_miner.sh
@@ -95,8 +95,8 @@ services:
 ENDDOCKERCOMPOSE
 
 cat > .env <<ENDENV
-SECRET_KEY=$(python3 -c 'import secrets; print(secrets.token_hex(25))')
-POSTGRES_PASSWORD=$(python3 -c 'import secrets; print(secrets.token_hex(8))')
+SECRET_KEY=$(python3 -c 'import secrets; print(secrets.token_urlsafe(25))')
+POSTGRES_PASSWORD=$(python3 -c 'import secrets; print(secrets.token_urlsafe(16))')
 BITTENSOR_NETUID=12
 BITTENSOR_NETWORK=finney
 BITTENSOR_WALLET_NAME="$(. ~/tmpvars && echo "$WALLET_NAME")"
--- a/install_miner.sh
+++ b/install_miner.sh
@@ -3,16 +3,16 @@ set -euxo pipefail
 
 if [ $# -ne 2 ]
 then
-  echo >2 "USAGE: ./install_miner.sh SSH_DESTINATION HOTKEY_PATH"
+  >&2 echo "USAGE: ./install_miner.sh SSH_DESTINATION HOTKEY_PATH"
   exit 1
 fi
 
 SSH_DESTINATION="$1"
 LOCAL_HOTKEY_PATH=$(realpath "$2")
-LOCAL_COLDKEY_PUB_PATH=$(dirname $(dirname "$LOCAL_HOTKEY_PATH"))/coldkeypub.txt
+LOCAL_COLDKEY_PUB_PATH=$(dirname "$(dirname "$LOCAL_HOTKEY_PATH")")/coldkeypub.txt
 
-if [ ! -f $LOCAL_HOTKEY_PATH ]; then
-  echo "Given HOTKEY_PATH does not exist"
+if [ ! -f "$LOCAL_HOTKEY_PATH" ]; then
+  >&2 echo "Given HOTKEY_PATH does not exist"
   exit 1
 fi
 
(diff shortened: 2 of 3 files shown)


install_validator.sh
```diff
@@ -0,0 +1,91 @@
...
+scp "$LOCAL_COLDKEY_PUB_PATH" "$SSH_DESTINATION:$REMOTE_COLDKEY_PUB_PATH"
+
+# install necessary software in the server
+ssh "$SSH_DESTINATION" <<'ENDSSH'
+set -euxo pipefail
+export DEBIAN_FRONTEND=noninteractive
+
+# install docker
+sudo apt-get update
+sudo apt-get install -y ca-certificates curl docker.io docker-compose
```
mjurbanski-reef, line 44: please install docker compose plugin

see:
https://github.com/reef-technologies/cookiecutter-rt-django/blob/ea836723a300070048df6ab9a7889fbdc6068406/%7B%7Bcookiecutter.repostory_name%7D%7D/bin/prepare-os.sh#L56

also lets keep using same docker distro i.e. either docker-ce should be used here or we need to change the one in cookiecutter if its bad for some reason

install_validator.sh
```diff
@@ -0,0 +1,91 @@
...
+BITTENSOR_NETUID=12
+BITTENSOR_NETWORK=finney
+BITTENSOR_WALLET_NAME="$(. ~/tmpvars && echo "$WALLET_NAME")"
+BITTENSOR_WALLET_HOTKEY_NAME="$(. ~/tmpvars && echo "$HOTKEY_NAME")"
+HOST_WALLET_DIR=$HOME/.bittensor/wallets
+FACILITATOR_URI=wss://facilitator.computehorde.io/ws/v0/
+ENDENV
+
+docker pull backenddevelopersltd/compute-horde-validator:v0-latest
+docker-compose up -d
```
mjurbanski-reef, line 89:
```suggestion
docker compose pull
docker compose up -d
```

install_miner.sh
```diff
@@ -0,0 +1,116 @@
...
+BITTENSOR_MINER_PORT=8000
+BITTENSOR_MINER_ADDRESS=auto
+PORT_FOR_EXECUTORS=8000
+ADDRESS_FOR_EXECUTORS=172.17.0.1
+ENDENV
+
+docker pull backenddevelopersltd/compute-horde-executor:v0-latest
+docker pull backenddevelopersltd/compute-horde-miner:v0-latest
+docker pull backenddevelopersltd/compute-horde-job:v0-latest
+docker-compose up -d
```
mjurbanski-reef, line 114:
```suggestion
docker compose pull
docker compose up -d
```

install_miner.sh
```diff
@@ -0,0 +1,116 @@
...
+      - /var/run/docker.sock:/var/run/docker.sock
+    command: --interval 60 --cleanup --label-enable
+ENDDOCKERCOMPOSE
+
+cat > .env <<ENDENV
+SECRET_KEY=$(python3 -c 'import secrets; print(secrets.token_hex(25))')
+POSTGRES_PASSWORD=$(python3 -c 'import secrets; print(secrets.token_hex(8))')
+BITTENSOR_NETUID=12
+BITTENSOR_NETWORK=finney
+BITTENSOR_WALLET_NAME="$(. ~/tmpvars && echo "$WALLET_NAME")"
```
mjurbanski-reef, line 102: this won't safely quote things that include `"` or `\`
what you want is `printf '%q' "$WALLET_NAME"` or `${WALLET_NAME@Q}`  https://unix.stackexchange.com/questions/379181/escape-a-variable-for-use-as-content-of-another-script

README.md
```diff
@@ -56,5 +56,20 @@ for miner in miners:
 
 # Running
 
-Refer to [Miner runner README](miner/envs/runner/README.md) and to 
-[Validator runner README](validator/envs/runner/README.md)
+To quickly start a validator or miner, create an ubuntu server and execute the following command from your local machine (where you have your wallet files).
```
mjurbanski-reef, line 59:
```suggestion
To quickly start a validator or miner, create an Ubuntu Server and execute the following command from your local machine (where you have your wallet files).
```

install_miner.sh
```diff
@@ -0,0 +1,116 @@
...
+    image: containrrr/watchtower:latest
+    restart: unless-stopped
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+    command: --interval 60 --cleanup --label-enable
+ENDDOCKERCOMPOSE
+
+cat > .env <<ENDENV
+SECRET_KEY=$(python3 -c 'import secrets; print(secrets.token_hex(25))')
+POSTGRES_PASSWORD=$(python3 -c 'import secrets; print(secrets.token_hex(8))')
```
mjurbanski-reef, line 99:
```suggestion
SECRET_KEY=$(python3 -c 'import secrets; print(secrets.token_urlsafe(25))')
POSTGRES_PASSWORD=$(python3 -c 'import secrets; print(secrets.token_urlsafe(16))')
```
lets make these passwords stronger

install_validator.sh
```diff
@@ -0,0 +1,91 @@
...
+    image: containrrr/watchtower:latest
+    restart: unless-stopped
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+    command: --interval 60 --cleanup --label-enable
+ENDDOCKERCOMPOSE
+
+cat > .env <<ENDENV
+SECRET_KEY=$(python3 -c 'import secrets; print(secrets.token_hex(25))')
+POSTGRES_PASSWORD=$(python3 -c 'import secrets; print(secrets.token_hex(8))')
```
mjurbanski-reef, line 79:
```suggestion
SECRET_KEY=$(python3 -c 'import secrets; print(secrets.token_urlsafe(25))')
POSTGRES_PASSWORD=$(python3 -c 'import secrets; print(secrets.token_urlsafe(16))')
```

mjurbanski-reef (approved):

install_validator.sh
emnoor-reef: `docker compose pull` will not pull these images. They are used by the compose file that is "inside" the runner.

I suppose pulling is not necessary here, because `up` will pull the runner, and runner's `up` will pull the validator image.

install_miner.sh
emnoor-reef: `docker compose pull` will not pull these images. They are used by the compose file that is "inside" the runner.

Pulling the miner is not necessary, but we **do need** to pull the executor and job images. Otherwise they will be pulled when the miner received a job and tries to execute it. (miner starts executor -> no image, pull; executor starts job -> no image, pull)

It would significantly increase the job duration (these images are huge) and negatively impact miner's weight.

install_validator.sh
emnoor-reef: I tried to keep them consistent with the runners [[1]](https://github.com/backend-developers-ltd/ComputeHorde/blob/master/miner/envs/runner/README.md), [[2]](https://github.com/backend-developers-ltd/ComputeHorde/blob/master/validator/envs/runner/README.md).

This project has not updated the cookiecutter template yet. I tried to do it once, but abandoned because of many people were in the middle of something (and still are) and it would cause a lot of conflicts (due to formatting changes in the cookiecutter).

I'd would prefer to update the whole project to compose v2 (as a separate task), than to make it inconsistent.

install_miner.sh
emnoor-reef: Never thought about that!! shell hard, me sad :(

I was doing some testing on how django-environ, docker and compose each interprets the `.env` files with unsafe values. They each interprets differently, even with peculier-but-"safe" values! 🤯 
Let me do some more digging, I'll post them later.

mjurbanski-reef: new docker compose inteprets .env files correctly, don't even check old docker-compose script since it is indeed broken:P

install_validator.sh
mjurbanski-reef: Lets not introduce new code that is already legacy at the point of of introduction.
If you think docker-compose command needs to be installed as well then do that, but lets make sure new code uses `docker compose` plugin. 

You won't need to change your yamls or anything AFAIK, it should just work.

mjurbanski-reef: (if you hit some problems with this, then sure, lets ignore it and continue as is, but IMO the barrier to make it right is so low that we waste more time but not doing it right away)

install_miner.sh
emnoor-reef: new docker compose may do, but python's environ lib doesn't.

we had a discussion on daily about this. even suggested that we "rename" the wallet path if it contains some special characters, so the env file is simple, and .env readers will be consistent.

PR #75, opened 2024-04-26
Update bittensor lib

andreea-popescu-reef:


compute_horde/pdm.lock
```diff
@@ -159,8 +159,11 @@ files = [
 
 [[package]]
 name = "bittensor"
-version = "6.8.1"
+version = "7.0.0"
 requires_python = ">=3.8"
+git = "https://github.com/opentensor/bittensor"
+ref = "merge_cuda"
```
mjurbanski-reef, line 165: please note this branch doesn't have bittensor 6.11.0 https://github.com/opentensor/bittensor/compare/merge_cuda...master changes

we may need to maintain our branch with these changes on our fork instead

mjurbanski-reef (approved):

compute_horde/pdm.lock
andreea-popescu-reef: yea could rebase the merge_cuda - but the current version has the hyperparam fetch fix so should be good for now

PR #77, opened 2024-04-26
miner runner updates

andreea-popescu-reef:


miner/envs/runner/_build-image.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -eux
```
mjurbanski-reef, line 1: https://github.com/reef-technologies/handbook/blob/master/docs/Code_conventions.md#safe-shell-settings
```suggestion
#!/bin/bash
set -eux -o pipefail
```
use of `-x` is script-context dependent, but rest are "sane defaults" 

please apply to other places as well (FYI  if you have any non-bash scripts as well - `sh` doesn't support `-o pipefail`)

miner/envs/runner/build-prod.sh
```diff
@@ -3,4 +3,5 @@
 ( cd nginx && ./build-image.sh )
 
 IMAGE_NAME="backenddevelopersltd/compute-horde-miner-runner:v0-latest"
-docker build --platform=linux/amd64 -t $IMAGE_NAME .
+MINER_IMAGE_REPO="compute-horde-miner"
+source _build-image.sh
```
mjurbanski-reef, line 7: https://github.com/reef-technologies/cookiecutter-rt-django/blob/master/%7B%7Bcookiecutter.repostory_name%7D%7D/bin/backup-db.sh#L3-L4
you may want to use SCRIPT_DIR= to be able to call this script without having to switch to SCRIPT_DIR directory yourself manually

miner/envs/runner/_publish-image.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -eux
+
+if [ -z "$(docker info 2>/dev/null | grep 'Username' | awk '{print $2}')" ]; then
```
mjurbanski-reef, line 3:
```suggestion
if [ -z "$(docker info 2>/dev/null | awk '/Username/ {print $2}')" ]; then
```
pretty sure you could also do `docker info --format  SOMETHING` but I guess it is not worth ($$$) figuring out what SOMETHING would be :P

mzukowski-reef (approved):

mjurbanski-reef (approved):
looks good, but I cannot review removal of `runner-staging` itself, so you may need to ask for second opinion on that one

miner/envs/runner/_build-image.sh
andreea-popescu-reef: fixed

PR #79, opened 2024-04-26
Miner version collection

emnoor-reef:


miner/envs/runner/nginx/build-image.sh
```diff
@@ -2,4 +2,4 @@
 set -eux -o pipefail
 
 IMAGE_NAME="backenddevelopersltd/compute-horde-miner-nginx:v0-latest"
-docker build -t $IMAGE_NAME .
+docker build --platform=linux/amd64 -t "$IMAGE_NAME" .
```
andreea-popescu-reef, line 5: why is this added?

andreea-popescu-reef (approved):

miner/envs/runner/nginx/build-image.sh
emnoor-reef: Well my machine is arm and I accidentally pushed an arm image before. So I add this to prevent myself from ruining other people's day again. :P

PR #80, opened 2024-04-26
Validator runner preprod

mzukowski-reef:


.github/workflows/miner_runner_cd_preprod.yml
```diff
@@ -0,0 +1,59 @@
+name: "Continuous Delivery: miner runner (prod)"
```
emnoor-reef, line 1:
```suggestion
name: "Continuous Delivery: miner runner (preprod)"
```

.github/workflows/miner_cd_preprod.yml
```diff
@@ -6,9 +6,9 @@ on:
       - 'deploy-miner-preprod'
 
 env:
-  TAG_VERSION: "v0-preprod"
+  TAG_VERSION: "v0-latest"
   STAGING_DOCKER_REPO_NAME: "backenddevelopersltd/compute-horde-miner-staging"
-  DOCKER_REPO_NAME: "backenddevelopersltd/compute-horde-miner"
+  DOCKER_REPO_NAME: "backenddevelopersltd/compute-horde-miner-preprod"
```
emnoor-reef, line 11: nitpick: since we are separating the docker repo for preprod, it would be nice to have one variable `MINER_IMAGE` instead of 2 passing as build args: `MINER_IMAGE_REPO` and `MINER_IMAGE_TAG` (and similar for validators).

not that important, but will make things "nice".

PR #83, opened 2024-04-30
Pass specs

andreea-popescu-reef:


validator/app/src/compute_horde_validator/validator/management/commands/connect_facilitator.py
```diff
@@ -219,6 +230,14 @@ def my_hotkey(self) -> str:
...
     async def run_forever(self) -> NoReturn:
         """ connect (and re-connect) to facilitator and keep reading messages ... forever """
+
+        # setup channel for receiving machine specs
+        self.channel_name = await self.channel_layer.new_channel()
+        await self.channel_layer.group_add(MACHINE_SPEC_GROUP_NAME, self.channel_name)
+
+
+        # send machine specs to facilitator
+        self.specs_task = asyncio.create_task(self.wait_for_specs())
```
emnoor-reef, line 240: `self.specs_task` is left running even if the use of facilitator client ends. It will be a problem when we finally start writing tests for this project! :-/

Clean up the task in `__aexit__` or somewhere.

emnoor-reef (approved):

emnoor-reef:
Please make the linter happy :)

validator/app/src/compute_horde_validator/validator/management/commands/connect_facilitator.py
andreea-popescu-reef: task is started from run_forever - so there's no cleanup point now - any tests would not use that anyway / or would refactor that whole part

PR #84, opened 2024-05-01
fix disk size scrapping

andreea-popescu-reef:


validator/app/src/compute_horde_validator/validator/synthetic_jobs/v1_decrypt.py
```diff
@@ -52,9 +52,9 @@ def scrape_specs() -> dict[str, any]:
         data['ram']['free'] = run_cmd('cat /proc/meminfo | grep -P "MemFree" | grep -o \"[0-9]*\"').strip()
         data['ram']['available'] = run_cmd('cat /proc/meminfo | grep -P "MemAvailable" | grep -o \"[0-9]*\"').strip()
         data['hard_disk'] = {}
-        data['hard_disk']['total'] = run_cmd('df . -P | sed -n 2p  | cut -d \' \' -f 9').strip()
-        data['hard_disk']['used'] = run_cmd('df . -P | sed -n 2p  | cut -d \' \' -f 10').strip()
-        data['hard_disk']['free'] = run_cmd('df . -P | sed -n 2p  | cut -d \' \' -f 13').strip()
+        data['hard_disk']['total'] = run_cmd('df . -P | sed -nr \'s/\\s+/ /g;2p\' | cut -d \' \' -f 2').strip()
+        data['hard_disk']['used'] = run_cmd('df . -P | sed -nr \'s/\\s+/ /g;2p\' | cut -d \' \' -f 3').strip()
+        data['hard_disk']['free'] = run_cmd('df . -P | sed -nr \'s/\\s+/ /g;2p\' | cut -d \' \' -f 4').strip()
```
mjurbanski-reef, line 60: tl;dr  please replace df with shutil.disk_usage(".") (just remember to divide by `// 1024` to the kB

----
I'm confused by this script, it seems to be spawning a lot more processes that it should - for example this second:
a) spawns `df` multiple times (with bash+sed+cut on top)
b) no extra processes are even necessary as `shutil.disk_usage` does the same job

Same with the all the cat&greps above.

the `MemUsed` doesn't even exists AFAIK and the types are mostly strings here (instead of int/float, but I guess pydantic fixes that on receiving side?)

my attempt at improving it slightly

```python
def run_cmd(cmd):
    try:
        return subprocess.check_output(cmd, shell=True).decode()
    except Exception:
        return ""


def scrape_specs() -> dict[str, any]:
    data: dict[str, any] = {}

    try:
        nvidia_cmd = run_cmd(
            "nvidia-smi --query-gpu=name,driver_version,name,memory.total,compute_cap,power.limit,clocks.gr,clocks.mem --format=csv"
        ).splitlines()
        csv_data = csv.reader(nvidia_cmd)
        header = [x.strip() for x in next(csv_data)]
        gpus = []
        for row in csv_data:
            row = [x.strip() for x in row]
            gpu_data = dict(zip(header, row))
            gpus.append(
                {
                    "name": gpu_data["name"],
                    "driver": gpu_data["driver_version"],
                    "capacity": gpu_data["memory.total [MiB]"].split(" ")[0],
                    "cuda": gpu_data["compute_cap"],
                    "power_limit": gpu_data["power.limit [W]"].split(" ")[0],
                    "graphics_speed": gpu_data["clocks.current.graphics [MHz]"].split(
                        " "
                    )[0],
                    "memory_speed": gpu_data["clocks.current.memory [MHz]"].split(" ")[
                        0
                    ],
                }
            )
        data["gpu"] = {"details": gpus, "count": len(gpus)}
    except Exception:
        # print(f'Error processing scraped gpu specs: {e}', flush=True)
        pass
    try:
        data["os"] = run_cmd(
            'lsb_release -d | grep -Po "Description:\\s*\\K.*"'
        ).strip()
        data["virtualization"] = run_cmd("virt-what").strip()

        data["ram"] = {}
        try:
            with open("/proc/meminfo") as f:
                meminfo = f.read()

            for name, key in [
                ("MemAvailable", "available"),
                ("MemFree", "free"),
                ("MemTotal", "total"),
            ]:
                data["ram"][key] = int(
                    re.search(rf"^{name}:\s*(\d+)\s+kB$", meminfo, re.M).group(1)
                )
            data["ram"]["used"] = data["ram"]["total"] - data["ram"]["free"]
        except Exception as e:
            print(f"Error reading /proc/meminfo; Exc: {e}", file=sys.stderr)

        disk_usage = shutil.disk_usage(".")
        data["hard_disk"] = {
            "total": disk_usage.total // 1024,
            "used": disk_usage.used // 1024,
            "free": disk_usage.free // 1024,
        }

        data["cpu"] = {}
        lscpu_output = run_cmd("lscpu")
        try:
            data["cpu"]["model"] = re.search(
                r"Model name:\s*(.*)$", lscpu_output, re.M
            ).group(1)
            data["cpu"]["count"] = int(
                re.search(r"CPU\(s\):\s*(.*)", lscpu_output).group(1)
            )
        except Exception:
            pass

        cpu_data = run_cmd('lscpu --parse=MHZ | grep -Po "^[0-9,.]*$"').splitlines()
        data["cpu"]["clocks"] = [float(x) for x in cpu_data]
    except Exception:
        # print(f'Error processing scraped specs: {e}', flush=True)
        raise
    return data
```

maybe I'm

mjurbanski-reef (approved):

PR #85, opened 2024-05-02
check miner status after install in installer script

emnoor-reef:


install_miner.sh
```diff
@@ -139,3 +139,33 @@ docker pull backenddevelopersltd/compute-horde-job:v0-latest
 docker compose up -d
 
 ENDSSH
+
+set +x
```
agoncharov-reef, line 143: But there is `set -euxo pipefail` in the beginning of the file?

install_miner.sh
```diff
@@ -139,3 +139,33 @@ docker pull backenddevelopersltd/compute-horde-job:v0-latest
...
+
+    echo "Miner installed successfully"
+    exit 0
+  fi
+
+  sleep "$run"
+done
+
+>&2 echo "Cannot connect to miner. Please check if everything is installed and running properly."
+>&2 echo "Also make sure port 8000 is reachable from outside (i.e. not blocked in firewall)."
```
agoncharov-reef, line 170: Wait, so smth is reachable from outside and it's not https?

install_miner.sh
emnoor-reef: Unsetting `-x` with `+x` here because we don't want to print the following commands, which are used to display pretty output.

install_miner.sh
emnoor-reef: Yes :)

PR #86, opened 2024-05-06
facilitator protocol documentation

emnoor-reef:


facilitator-protocol.md
```diff
@@ -0,0 +1,129 @@
...
+  "message_type": "V0AuthenticationRequest",
+  "public_key": "...",
+  "signature": "..."
+}
+```
+
+| Field        | Details                                                                                               |
+|--------------|-------------------------------------------------------------------------------------------------------|
+| `public_key` | the hex of the public key of validator hotkey                                                         |
+| `signature`  | the hex of the signature of the validator hotkey's public key with the private key prefixed with `0x` |
```
mzukowski-reef, line 47: this is "authentication" :) we cannot release protocol with it :)

in my PoC of mTLS, authentication was possible using "static" signature because TLS was providing a proof that person has access to private key of TLS certificate - so signature was saying: public key of this cert was signed with private hotkey. But still I was not happy with level of security - you may "trick" person to sign a some string... that is why we are trying to add an encryption key to axon info... 

this implementation is open to "chosen-plaintext attack", but also one can somehow sniff communication and read it once and then authorize...

this is not how this should be done

mzukowski-reef: something like JWT token with short expiration date would be acceptable

facilitator-protocol.md
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+```
+
+| Field                   | Details                                               |
+|-------------------------|-------------------------------------------------------|
+| uuid                    | the unique job uuid this status is for                |
+| status                  | `failed`, `rejected`, `accepted` or, `completed`      |
+| metadata                | (optional) the metadata for this status               |
+| metadata.comment        | summery of the metadata                               |
```
mzukowski-reef, line 106:
```suggestion
| metadata.comment        | summary of the metadata                               |
```

facilitator-protocol.md
```diff
@@ -0,0 +1,129 @@
...
+| metadata.miner_response | (optional) the response of miner when a job completes |
```
mzukowski-reef, line 107: shouldn't structure of miner response be documented?

mzukowski-reef (changes requested):

mzukowski-reef (commented):
Also I've added Heartbeat request - validator sends simple heatbeat request every minute

mzukowski-reef (changes requested):
also sending machine specs is implemented in such way that it is working for all validators and you can't turn it off :) so... we have to document it

facilitator-protocol.md
emnoor-reef: Here, I'm just documenting what is already implemented. I think it needs to be discussed as part of the protocol design.

facilitator-protocol.md
emnoor-reef: I'll add it. I left it out, as miner response is part of miner-validator protocol.

emnoor-reef:
I'll add the heartbeat request.

I left out the machine specs as per this discussion: https://reeftechnologies.slack.com/archives/C06DMMT987M/p1715008514007749

PR #89, opened 2024-05-08
/dev/urandom password generator

adalc-reef:
50% or so faster

### DIFF
--- a/validator/app/src/compute_horde_validator/validator/synthetic_jobs/v1_synthetic_job.py
+++ b/validator/app/src/compute_horde_validator/validator/synthetic_jobs/v1_synthetic_job.py
@@ -29,9 +29,8 @@ def gen_keys(num_keys, num_letters, num_digits):
     alphabet = string.ascii_letters + string.digits
     alphabet_len = len(alphabet)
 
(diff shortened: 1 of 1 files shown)


validator/app/src/compute_horde_validator/validator/synthetic_jobs/v1_synthetic_job.py
```diff
@@ -18,6 +18,35 @@
...
+        marker = new_marker
+
+        chars = [' '] * key_len
+        for i in range(num_letters):
+            number, r = divmod(number, alphabet_len)
+            chars[i] = alphabet[r]
+        for i in range(num_digits):
+            number, r = divmod(number, digits_len)
+            chars[num_letters + i] = digits[r]
+        yield ''.join(chars)
```
mjurbanski-reef, line 48: all this weirdness could be avoided just by using `?h` mask instead of digits and letters seperately https://hashcat.net/wiki/doku.php?id=mask_attack instead of letters and digits separately - there is really nothing to gain by current approach AFAIK.

After the change for the "password" generation you just call `secrets.token_hex(n)` and you are done without any of this weirdness.

PR #91, opened 2024-05-09
Job receipts

emnoor-reef:


miner/app/src/compute_horde_miner/miner/miner_consumer/validator_interface.py
```diff
@@ -182,6 +195,29 @@ async def handle(self, msg: BaseValidatorRequest):
             job.full_job_details = msg.dict()
             await job.asave()
 
+        if isinstance(msg, validator_requests.V0ReceiptRequest) and self.verify_receipt_msg(msg):
+            logger.debug(f'Received receipt for job_uuid: {msg.payload.job_uuid}')
```
pawelpolewicz, line 199: should be INFO and should contain validator hotkey, time_took and score so that people can grep and parse it out from the logs easily

miner/app/src/compute_horde_miner/miner/utils.py
```diff
@@ -0,0 +1,16 @@
...
+
+from compute_horde_miner.miner.models import JobReceipt
+from compute_horde_miner.miner.receipt_store.base import Receipt
+from compute_horde_miner.miner.receipt_store.current import receipts_store
+
+RECEIPTS_MAX_RETENTION_PERIOD = datetime.timedelta(days=7)
+
+
+async def prepare_receipts() -> None:
+    # TODO: use a lock to prevent simultaneous write to zip file
```
mzukowski-reef, line 13: this TODO is misleading - current implementation is asyncio blocking, so we actually have this lock :D but even then this code is wrong - you can run this after two new receipts and older query would race with newer query in that way that old receipts that don't have newst object would overwrite a newer snapshot (it would be fixed after next recepit, but still)

miner/app/src/compute_horde_miner/miner/receipt_store/local.py
```diff
@@ -0,0 +1,25 @@
...
+class LocalReceiptStore(BaseReceiptStore):
+    async def store(self, receipts: list[Receipt]):
+        if not receipts:
+            return
+
+        root = pathlib.Path(settings.LOCAL_RECEIPTS_ROOT)
+        root.mkdir(parents=True, exist_ok=True)
+        filepath = root / ZIP_FILENAME
+
+        with zipfile.ZipFile(filepath, mode='w', compression=zipfile.ZIP_LZMA) as zf:
```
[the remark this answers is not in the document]
mzukowski-reef: I know that the design was "make it zip" - but I still give myself right to complain :P

miner/app/src/compute_horde_miner/miner/models.py
```diff
@@ -70,3 +75,20 @@ async def get_not_reported(cls, validator: Validator) -> Iterable[Self]:
...
+    validator_signature = models.CharField(max_length=256)
+    miner_signature = models.CharField(max_length=256)
+
+    # payload fields
+    job_uuid = models.UUIDField()
+    miner_hotkey = models.CharField(max_length=256)
+    validator_hotkey = models.CharField(max_length=256)
+    time_started = models.DateTimeField()
+    time_took = models.DurationField()
+    score = models.FloatField()
```
mzukowski-reef, line 91: I am concerned about "float stability" - we generate signature of float dumped to string - and then we post this over network, json loads, then pydantic parse to float, then check signature - then store in database, then read from database and dumps to json...  same with DurationField - it is float in json - don't know what it is in database - DateTime? also not sure...

miner/app/src/compute_horde_miner/miner/receipt_store/base.py
```diff
@@ -0,0 +1,43 @@
...
+    payload: ReceiptPayload
+    validator_signature: str
+    miner_signature: str
+
+    @classmethod
+    def from_job_receipt(cls, jr: JobReceipt) -> Self:
+        return cls(
+            payload=ReceiptPayload(
+                job_uuid=str(jr.job_uuid),
+                miner_hotkey=settings.BITTENSOR_WALLET().hotkey.ss58_address,
```
mzukowski-reef, line 21: miner can change key without changing database (I think) - then we actually should be filtering out old receipts

mzukowski-reef (changes requested):

miner/app/src/compute_horde_miner/miner/utils.py
```diff
@@ -0,0 +1,16 @@
...
+    await JobReceipt.objects.filter(time_started__lt=now()-RECEIPTS_MAX_RETENTION_PERIOD).adelete()
```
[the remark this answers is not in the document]
emnoor-reef: The deletes? Or the whole `prepare_receipts` thing?

even wasn't keen on beat for the zip generation: https://reeftechnologies.slack.com/archives/C06DMMT987M/p1715245099112639?thread_ts=1715244515.728119&cid=C06DMMT987M
and I was struggling to make it a normal celery task (conflicts with asyncio).

I'll move the deletions to a beat task.

miner/app/src/compute_horde_miner/miner/utils.py
```diff
@@ -0,0 +1,13 @@
...
+
+from compute_horde_miner.miner.models import JobReceipt
+from compute_horde_miner.miner.receipt_store.base import Receipt
+from compute_horde_miner.miner.receipt_store.current import receipts_store
+
+_prepare_receipts_lock = asyncio.Lock()
+
+
+async def prepare_receipts() -> None:
+    async with _prepare_receipts_lock:
```
mzukowski-reef, line 11: this only make sense if you call it from the same asyncio loop - but you call it from celery task and django command - by default celery runs concurrent tasks in procesess.. and also in both places you run it in sync code so there is no reason for async here

mzukowski-reef (changes requested):

PR #95, opened 2024-05-16
Scraping errors: return stderr of failed command

adalc-reef:
Return the output of failed commands from specs scrape in synthetic job

### DIFF
--- a/validator/app/src/compute_horde_validator/validator/synthetic_jobs/v1_decrypt.py
+++ b/validator/app/src/compute_horde_validator/validator/synthetic_jobs/v1_decrypt.py
@@ -11,14 +11,12 @@ from cryptography.fernet import Fernet
 
 
 def run_cmd(cmd):
-    proc = subprocess.run(cmd, shell=True, capture_output=True, check=False)
-    stdout = proc.stdout.decode()
+    proc = subprocess.run(cmd, shell=True, capture_output=True, check=False, text=True)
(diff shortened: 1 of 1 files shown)


validator/app/src/compute_horde_validator/validator/synthetic_jobs/v1_decrypt.py
```diff
@@ -11,10 +11,15 @@
...
-    except Exception:
-        return ""
+    proc = subprocess.run(cmd, shell=True, capture_output=True, check=False)
+    stdout = proc.stdout.decode()
+    if proc.returncode != 0:
+        stderr = proc.stderr.decode()
+        raise RuntimeError(
+            f"[errno {proc.returncode}] {cmd}\nSTDOUT:\n{stdout}\nSTDERR:\n{stderr}"
+        )
+    return stdout
```
mjurbanski-reef, line 21: suggestion:
a) text=True takes care of decoding for you
b) repr escapes output properly
```suggestion
    proc = subprocess.run(cmd, shell=True, capture_output=True, check=False, text=True)
    if proc.returncode != 0:
        raise RuntimeError(
            f"run_cmd error {cmd=!r} {proc.returncode=} {proc.stdout=!r} {proc.stderr=!r}"
        )
    return proc.stdout
```

mjurbanski-reef (approved):

PR #102, opened 2024-05-21
fix nox lint

emnoor-reef:


miner/noxfile.py
```diff
@@ -126,7 +126,7 @@ def lint(session):
     session.run("codespell", ".", "--skip='*.lock'")
     run_shellcheck(session, mode="check")
     run_readable(session, mode="check")
-    session.run("ruff", "format", ".")
+    session.run("ruff", "check", "--diff", "--unsafe-fixes", ".")
```
mjurbanski-reef, line 129: ruff check
and
ruff format

 have completely different function

mjurbanski-reef (changes requested):

miner/noxfile.py
mjurbanski-reef: fixed this in https://github.com/backend-developers-ltd/ComputeHorde/pull/103

PR #105, opened 2024-05-22
Scrape hardware stats from the executor

adalc-reef:
Scrape stats by running an `nvidia-smi` container from the executor, not from inside the synthetic job container.

Synthetic job container will continue to do the scraping, but we'll ignore the results. Will remove in further PR to keep changes minimal.

### DIFF
--- a/executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
+++ b/executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
@@ -392,8 +392,6 @@ class JobRunner:
             exit_status = process.returncode
             timeout = False
 
-        specs = get_machine_specs()
-
         # Save the streams in output volume and truncate them in response.
         with open(self.output_volume_mount_dir / "stdout.txt", "w") as f:
             f.write(stdout)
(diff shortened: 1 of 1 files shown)


executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -301,11 +392,7 @@ async def run_job(self, job_request: V0JobRequest):
             exit_status = process.returncode
             timeout = False
 
-        # fetch machine specs if synthetic job generated
-        specs = None
-        if (self.specs_volume_mount_dir / "specs.json").exists():
-            with open(self.specs_volume_mount_dir / "specs.json") as f:
-                specs = MachineSpecs(specs=json.load(f))
+        specs = get_machine_specs()
```
mzukowski-reef, line 395: can we refactor it a bit so that it is called before `await self.miner_client.send_ready()` in _executor_loop - so it gathering this data does not impact on overal score?

mzukowski-reef (changes requested):

PR #112, opened 2024-05-24
Support batch_id in miner specs update

mzukowski-reef:


validator/app/src/compute_horde_validator/validator/synthetic_jobs/utils.py
```diff
@@ -364,6 +371,7 @@ async def _execute_job(
                 {
                     "type": "machine.specs",
                     "miner_hotkey": job.miner.hotkey,
+                    "batch_id": str(batch_id_to_uuid(job.batch_id)),
```
mjurbanski-reef, line 374: you need to convert batch_id to get a batch_id ? thats kinda confusing

maybe this should be called 
```suggestion
                    "batch_uuid": str(batch_id_to_uuid(job.batch_id)),
```
then?

mzukowski-reef: I pointed this in facilitator review, but it got merged and this is the interface :D

PR #117, opened 2024-05-28
Fix debug_run_synthetic_job_to_miner

adalc-reef:


validator/app/src/compute_horde_validator/validator/management/commands/debug_run_syntetic_job_to_miner.py
```diff
@@ -1,11 +1,17 @@
 import asyncio
 import datetime
```
andreea-popescu-reef, line 2: you could also fix debug_run_synthetic_jobs while your at it - should be similat

PR #123, opened 2024-06-05
Allow executors that timeouts on barrier to executue job

mzukowski-reef:


validator/app/src/compute_horde_validator/validator/synthetic_jobs/utils.py
```diff
@@ -44,6 +44,7 @@
 JOB_LENGTH = 300
 TIMEOUT_LEEWAY = 1
 TIMEOUT_MARGIN = 60
+TIMEOUT_BARRIER = JOB_LENGTH - 65
```
emnoor-reef, line 47: nitpick: why was `65` chosen here? maybe a named constant would be clearer.

emnoor-reef (approved):

PR #125, opened 2024-06-06
spec sending refactor logs

andreea-popescu-reef:
added logs - small refactor

### DIFF
--- a/validator/app/src/compute_horde_validator/validator/facilitator_client.py
+++ b/validator/app/src/compute_horde_validator/validator/facilitator_client.py
@@ -160,15 +160,18 @@ class FacilitatorClient:
                 try:
                     await self.handle_connection(ws)
                 except websockets.ConnectionClosed as exc:
+                    self.ws = None
                     logger.warning(
                         "validator connection closed with code %r and reason %r, reconnecting...",
                         exc.code,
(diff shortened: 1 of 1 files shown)


validator/app/src/compute_horde_validator/validator/facilitator_client.py
```diff
@@ -194,29 +197,39 @@ async def wait_for_specs(self):
...
+                    while len(specs_queue) > 0:
+                        spec_to_send = specs_queue.pop(0)
+                        try:
+                            await self.send_model(spec_to_send)
+                        except Exception as exc:
+                            logger.error("Error occurred while sending specs", exc_info=exc)
+                            specs_queue.insert(0, spec_to_send)
+                            break
+            except TimeoutError:
+                logger.debug("wait_for_specs still running")
```
emnoor-reef, line 224: I think this log is a bit confusing.

```suggestion
                logger.debug("Timed out while waiting for machine specs after {X} seconds. Trying again...")
```

andreea-popescu-reef: added the timeout just to verify that the tasks still runs to see if the bug re-appears - will remove it after I confirm is fixed

PR #129, opened 2024-06-11
change receipts format to csv

emnoor-reef:


miner/app/src/compute_horde_miner/miner/receipt_store/local.py
```diff
@@ -17,15 +18,39 @@ def store(self, receipts: list[Receipt]) -> None:
...
+                        receipt.miner_signature,
+                    ]
+                )
 
         shutil.move(temp_file.name, filepath)
         filepath.chmod(0o644)
 
     def get_url(self) -> str:
-        return settings.LOCAL_RECEIPTS_URL + ZIP_FILENAME
+        return settings.LOCAL_RECEIPTS_URL + FILENAME
```
mzukowski-reef, line 56: I think you removed redirect view so we don't use it so it can be removed - but it is suggestion to consider :)

mzukowski-reef (approved):

PR #131, opened 2024-06-12
Allow changing synthetic job schedule for debug

mzukowski-reef:


emnoor-reef (commented):
Doesn't the existing `DEBUG_RUN_BEAT_VERY_OFTEN` env solve this already, if we just tweak the values a bit?

emnoor-reef (approved):

PR #141, opened 2024-06-26
Added executor_class to most Job models/requests

adalc-reef:


mzukowski-reef (approved):
you can change in miner

await current.executor_manager.reserve_executor(token)

to use:
async def reserve_executor_class(self, token, executor_class, timeout):

and it should work - but it can be added later with all dedicated synthetic jobs and stuff

PR #154, opened 2024-07-02
Raise exceptions returned from asyncio.gather()

emnoor-reef:


validator/app/src/compute_horde_validator/validator/synthetic_jobs/utils.py
```diff
@@ -353,6 +359,15 @@ async def execute_miner_synthetic_jobs(batch_id, miner_id, miner_hotkey, axon_in
             miner_client.add_job(str(job.job_uuid))
         try:
             await execute_synthetic_jobs(miner_client, jobs)
+        except ExceptionGroup as e:
+            msg = f"Multiple errors occurred during execution of some jobs for miner {miner_hotkey}"
+            logger.warning(f"{msg}: {e!r}")
+            stacktrace = "".join(traceback.format_exception(e))
+            long_description = msg + "\n" + stacktrace
+            await save_job_execution_event(
+                subtype=SystemEvent.EventSubType.MULTIPLE_ERRORS_DURING_JOB,
```
andreea-popescu-reef, line 368: not sure if we want to system event it? these seem like the type of errors we want to get in sentry and investigate / handle separately or sys event them as GENERIC_ERROR

andreea-popescu-reef (approved):

PR #156, opened 2024-07-02
Inject validator for local development bypassing fetching

olzhasar-reef:
There is already a `debug_add_validator` command in the miner that can be used to add a validator during the local development. However, this validator will be deactivated once a user starts celery beat (a periodic job fetches validators from the metagraph and deactivates records not present there).
This PR introduces a new `DEBUG_VALIDATOR_KEY` setting (read from an env var) allowing to bypass this mechanism.


miner/app/src/compute_horde_miner/miner/tests/test_tasks.py
```diff
@@ -0,0 +1,91 @@
...
+    inactive_validators: list[Validator],
+    mock_get_validators: MagicMock,
+):
+    mock_get_validators.return_value = _generate_validator_mocks(inactive_validators_keys)
+
+    fetch_validators()
+
+    assert Validator.objects.count() == len(active_validators) + len(inactive_validators)
+
+    assert Validator.objects.filter(active=True).count() == len(inactive_validators)
```
emnoor-reef, line 70: suggestion: change the counts of active and inactive keys in the fixture, so that this check is more effective.

miner/app/src/compute_horde_miner/miner/tests/test_tasks.py
```diff
@@ -0,0 +1,91 @@
...
+
+    for key in inactive_validators_keys:
+        assert Validator.objects.get(public_key=key).active is True
```
emnoor-reef, line 73:
```suggestion
        assert Validator.objects.get(public_key=key).active
```

emnoor-reef (approved):

miner/app/src/compute_horde_miner/miner/tests/test_tasks.py
olzhasar-reef: I've switched to an `exists()` query here to unify checking style with a test case before

emnoor-reef (approved):

emnoor-reef (approved):

PR #157, opened 2024-07-03
add job started receipt

andreea-popescu-reef:


validator/app/src/compute_horde_validator/validator/synthetic_jobs/utils.py
```diff
@@ -454,6 +472,24 @@ async def _execute_synthetic_job(miner_client: MinerClient, job: SyntheticJob):
         await save_event(subtype=SystemEvent.EventSubType.FAILURE, long_description=comment)
         raise ValueError(comment)
 
+    # Send job started receipt to miner
+    try:
+        receipt_message = miner_client.generate_job_started_receipt_message(
+            job,
+            time.time(),
+            settings.MAX_SYNTHETIC_JOB_TIMEOUT,
```
emnoor-reef, line 480: Shouldn't this be `job_generator.timeout_seconds()` instead of a static value from settings? Moreover, different executor classes have different timeouts, so I don't know if introducing this `MAX_SYNTHETIC_JOB_TIMEOUT` settings is the right move.

validator/app/src/compute_horde_validator/settings.py
```diff
@@ -363,6 +363,7 @@ def wrapped(*args, **kwargs):
 
 MANIFEST_SCORE_MULTIPLIER = 1.05
 EXECUTOR_COUNT_INCREASE_THRESHOLD = 3
+MAX_SYNTHETIC_JOB_TIMEOUT = 60  # seconds
```
emnoor-reef, line 366: I don't understand the purpose of this option TBH. (see my other comment on `utils.py`.

emnoor-reef (changes requested):

validator/app/src/compute_horde_validator/validator/synthetic_jobs/utils.py
andreea-popescu-reef: can't be just `job_generator.timeout_seconds()` as it also includes some sending the task preparing exeutor etc - but you're right it should depend on the job_generator.timeout_seconds
changed it but exact values need to be determined

miner/app/src/compute_horde_miner/miner/tasks.py
```diff
@@ -57,7 +57,7 @@ def fetch_validators():
 
 @app.task
 def prepare_receipts():
-    job_receipts = JobReceipt.objects.order_by("time_started").filter(
+    job_receipts = JobFinishedReceipt.objects.order_by("time_started").filter(
```
mzukowski-reef, line 60: we are not serving JobStartedReceipt?

miner/app/src/compute_horde_miner/miner/tasks.py
```diff
@@ -84,13 +89,13 @@ def get_receipts_from_old_miner():
     receipts = get_miner_receipts(hotkey, config.OLD_MINER_IP, config.OLD_MINER_PORT)
 
     latest_job_receipt = (
-        JobReceipt.objects.filter(miner_hotkey=hotkey).order_by("-time_started").first()
+        JobFinishedReceipt.objects.filter(miner_hotkey=hotkey).order_by("-time_started").first()
```
mzukowski-reef, line 92: Should we also migrate JobStartedRecepts? or it is not needed?

mzukowski-reef (commented):
after my changes in test_utils are merged we could easily add tests for checking if start receipts are sent

miner/app/src/compute_horde_miner/miner/tasks.py
emnoor-reef: I don't think serving them are necessary. The only purpose these receipts serve right now is to prove a job is running while rejecting other jobs.

miner/app/src/compute_horde_miner/miner/tasks.py
emnoor-reef: I could be wrong, but I don't think miners can keep their running jobs running while migrating, so they don't have to show these receipts for rejecting other jobs.

miner/app/src/compute_horde_miner/miner/tasks.py
```diff
@@ -62,16 +66,29 @@ def fetch_validators():
...
+        time_accepted__gt=now() - RECEIPTS_MAX_SERVED_PERIOD
+    )
+    receipts += [jr.to_receipt() for jr in job_started_receipts]
+
+    job_finished_receipts = JobFinishedReceipt.objects.order_by("time_started").filter(
         time_started__gt=now() - RECEIPTS_MAX_SERVED_PERIOD
     )
-    receipts = [jr.to_receipt() for jr in job_receipts]
+    receipts += [jr.to_receipt() for jr in job_finished_receipts]
+
```
mzukowski-reef, line 80: maybe we should sort whole collection by time_accepted | time_started?

validator/app/src/compute_horde_validator/validator/tests/test_utils.py
```diff
@@ -20,7 +20,7 @@
     V0JobFinishedRequest,
 )
 from compute_horde.mv_protocol.validator_requests import (
-    V0ReceiptRequest,
+    V0JobFinishedReceiptRequest,
```
mzukowski-reef, line 23: we don't have tests that checks if receipts are sent - this line happened because I am testing "manifest dance" with checking receipts - but please add asserts in test_execute_synthetic_job that checks both finished and started receipts

miner/app/src/compute_horde_miner/miner/receipt_store/local.py
```diff
@@ -16,38 +21,43 @@ def store(self, receipts: list[Receipt]) -> None:
         if not receipts:
             return
 
+        payload_fields = set()
```
mzukowski-reef, line 24: this is completly uncovered with tests

mzukowski-reef (changes requested):

PR #158, opened 2024-07-03
Fix double migrate in validator

emnoor-reef:


emnoor-reef:
`app` is not essential for validators. It wasn't in the compose file before, and it could be removed in the future. That's why I kept the one in celery entrypoint.

PR #175, opened 2024-07-16
Sign

andreea-popescu-reef:


.github/workflows/end_to_end_test_ci.yml
```diff
@@ -21,6 +21,8 @@ env:
   MNEMONIC_VALIDATOR_HOTKEY: "syrup ill organ rigid supreme pen menu range tonight letter pear blind"
   MINER_HOTKEY_SS58: "5FqSEa5kHZgwKMMuC7WsHLoXcf3UN9gCpDZ7iDifa2sTLBSB"
   VALIDATOR_HOTKEY_SS58: "5GHikjqiHD1pUyCf3mzrjbEJgJvGnCp5FNKTXTtGkzhaJ37R"
+  DOCKER_CONTENT_TRUST: 1
+  DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE: "not_relevant"
```
andreea-popescu-reef, line 25: this is only used to decrypt the delegation key (which itself is a github secret)

emnoor-reef: I don't quite understand why it is here

.github/workflows/end_to_end_test_ci.yml
emnoor-reef, line 1: I don't think we need to sign the e2e images

executor/app/src/compute_horde_executor/settings.py
```diff
@@ -43,6 +43,9 @@ def wrapped(*args, **kwargs):
 
 ENV = env("ENV", default="prod")
 
+DOCKER_CONTENT_TRUST = 1
```
emnoor-reef, line 46: Why is it needed here?

miner/app/src/compute_horde_miner/settings.py
```diff
@@ -48,6 +48,10 @@ def wrapped(*args, **kwargs):
 
 ENV = env("ENV", default="prod")
 
+
+DOCKER_CONTENT_TRUST = 1
```
emnoor-reef, line 52: Why is it needed here?

validator/app/Dockerfile
emnoor-reef, line 1: Isn't this symlink created with `setup-prod.sh`? I think this should not be committed.

validator/app/src/entrypoint.sh
emnoor-reef, line 1: Isn't this symlink created with setup-prod.sh? I think this should not be committed.

scripts/dct_pull/dct_pull.sh
```diff
@@ -0,0 +1,21 @@
...
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m'
+
+while true; do
+    for image in "$@"; do 
+        echo -e "\n\nPulling $image"
+        docker pull $image && echo -e "${GREEN}Successfully pulled signed image: $image${NC}" || echo -e "${RED}Failed to pull signed image: $image${NC}"
+    done
+    sleep 5m
```
emnoor-reef, line 19:
```suggestion
    sleep 1m
```

emnoor-reef (approved):

.github/workflows/end_to_end_test_ci.yml
andreea-popescu-reef: it's there to not get prompted for repository key when first pushing

executor/app/src/compute_horde_executor/settings.py
andreea-popescu-reef: not needed

PR #185, opened 2024-07-30
COM-120 commit reveal weights

agoncharov-reef:
This PR adds functionality to commit & (later) reveal weights instead of simple setting weights.
Here's a description from BT team: https://docs.bittensor.com/subnets/commit-reveal

New workflow:
0. `do_set_weights` celery task reads `commit_reveal_weights_enabled` hyperparameter
1. if `commit_reveal_weights_enabled` is `True`:
  - the task commits the hashed weights to the network and stores payload in `Weights` model instance
  - the task `reveal_scores` is run each minute and searched for not-yet-revealed `Weights` which are older than `commit_reveal_weights_interval - reveal_in_advance_num_blocks`, and reveals them
2. if  `commit_reveal_weights_enabled` is `False`, ol' good `set_weights` is called which sets weights immediately and they become available to everyone
3. if metagraph sucks and it's impossible to get value of `commit_reveal_weights_enabled`, noth `commit_weights` and `set_weights` are called to see which one will work

# Commit reveal interval
I tried to experiment with `commit_reveal_weights_interval`. Docs say:
> Specify the desired number of blocks of delay before revealing the weights, by using the new subnet hyperparameter commit_reveal_weights_interval (integer).

It is expected that weights should be revealed any time _after_ this interval, however during local testing I was able to reveal weights before that period (which is surprising to me) and some time after that period has passed. If revealing too late, the network will throw an error. I haven't figured out what are the exact rules, so we start trying to reveal weights a bit in advance (`reveal_in_advance_num_blocks: int = 10`).

# Weights are ints!
Also I noticed that in order to hash weights before committing them, the weight values must be `int`s. It was a surprise to me cause I thought they could be float, so if they can indeed then let's discuss wtf is going on.

# How to activate (or deactivate)
To enable the feature, one should modify subnet hyperparams:
```
btcli sudo set hyperparameters --netuid=xx --param commit_reveal_weights_enabled --value True --wallet.name local-owner
btcli sudo set hyperparameters --netuid=12 --param commit_reveal_weights_interval --value 20 --wallet.name local-owner
```

# Is this bs tested?
Yes, tested on local bittensor network.


validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -388,6 +434,77 @@ def set_scores():
...
+@app.task()
+def reveal_scores(reveal_in_advance_num_blocks: int = 10) -> None:
+    """
+    Selects Weights that are older than `commit_reveal_weights_interval`
+    and haven't been revealed yet, and reveal them.
+    """
+    subtensor = get_subtensor(network=settings.BITTENSOR_NETWORK)
+
+    try:
+        hyperparams = subtensor.get_subnet_hyperparameters(netuid=settings.BITTENSOR_NETUID)
```
mzukowski-reef, line 446: don't relay on hyperparam here - if there is unrevealed weights object we should try to reveal it - if commit_reveal_weights_enabled is not enabled then when creating weights in db we should mark them as revealed

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -388,6 +434,77 @@ def set_scores():
...
+        if not hyperparams.commit_reveal_weights_enabled:
+            logger.debug("Commit reveal weights is not enabled")
+            return
+
+        commit_reveal_weights_interval = hyperparams.commit_reveal_weights_interval
+    except Exception:
+        logger.exception('Failed to fetch "commit_reveal_weights_enabled" hyperparameter')
+        commit_reveal_weights_interval = 0
```
mzukowski-reef, line 454: we should have better default

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -388,6 +434,77 @@ def set_scores():
...
+
+    unrevealed_weights = (
+        Weights.objects
+        .filter(
+            revealed_at=None,
+            block__lte=current_block_number - (commit_reveal_weights_interval - reveal_in_advance_num_blocks),
+        )
+        .order_by("created_at")
+    )
+    for weights in unrevealed_weights:
```
mzukowski-reef, line 467: I think that @ppolewicz said that you can reveal only the last commited weights - if I heard correctly then we should prevent commiting weights if previous one are no expired and still unrevealed - and then in this place we would reveal only latest weights

mzukowski-reef (changes requested):

validator/app/src/compute_horde_validator/validator/tasks.py
agoncharov-reef: Yes, makes sense, thank you!
Although the second part is not true - we cannot create `Weights` when `commit_reveal_weights_enabled` is not enabled .

validator/app/src/compute_horde_validator/validator/tasks.py
agoncharov-reef: The idea was: if we have no fucking clue what's the interval because of stupid broken metagraph or so, then fallback to "almost regular set-weight" == "commit weights and reveal as soon as possible". When metagraph is finally fixed, return back to regular interval. As far as I understand, when we introduce commit-reveal, cheaters won't try to eliminate it, and even if we fallback to zero interval for some moment, nobody will use it.

validator/app/src/compute_horde_validator/validator/tasks.py
agoncharov-reef: Thank you, good point!

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -388,6 +459,66 @@ def set_scores():
...
+            netuid=settings.BITTENSOR_NETUID,
+            uids=weights.uids,
+            weights=weights.weights,
+            salt=weights.salt,
+            version_key=weights.version_key,
+            wait_for_inclusion=True,
+            wait_for_finalization=True,
+            max_retries=2,
+        )
+        if not is_success:
```
mzukowski-reef, line 518: we should have some weight reveal system events - also we might think to send different event metadata for weight commit than weight setting

mzukowski-reef (changes requested):
we cant merge with failing tests

PR #190, opened 2024-08-03
Normlize weights before committing them

mpnowacki-reef:


validator/app/src/compute_horde_validator/validator/tests/helpers.py
```diff
@@ -216,14 +221,19 @@ def set_weights(
...
+            weights = weights.tolist()
+        self.weights_set.append(weights)
         return self.mocked_set_weights()
 
-    def commit_weights(self, **kwargs) -> tuple[bool, str]:
+    def commit_weights(self, weights, **kwargs) -> tuple[bool, str]:
+        self.weights_committed.append(weights)
         if self.hyperparameters.commit_reveal_weights_enabled:
             return True, ""
         return False, "MockSubtensor doesn't support commit_weights"
```
agoncharov-reef, line 233: I think this is more correct
```python
if self.hyperparameters.commit_reveal_weights_enabled:
    self.weights_committed.append(weights)
    return True, ""
return False, "MockSubtensor doesn't support commit_weights"
```

validator/app/src/compute_horde_validator/validator/tests/helpers.py
```diff
@@ -216,14 +221,19 @@ def set_weights(
...
 
-    def reveal_weights(self, **kwargs) -> tuple[bool, str]:
+    def reveal_weights(self, weights, **kwargs) -> tuple[bool, str]:
+        self.weights_revealed.append(weights)
         if self.hyperparameters.commit_reveal_weights_enabled:
             return True, ""
         return False, "MockSubtensor doesn't support reveal_weights"
```
agoncharov-reef, line 239:
```python
if self.hyperparameters.commit_reveal_weights_enabled:
    self.weights_revealed.append(weights)
    return True, ""
return False, "MockSubtensor doesn't support reveal_weights"
```

agoncharov-reef (approved):

validator/app/src/compute_horde_validator/validator/tests/helpers.py
mpnowacki-reef: yeah but then if we want to set what values were passed to this call we won't be able to

PR #193, opened 2024-08-06
Fix bare except

adalc-reef:


bswck:
There is a lot to fix out there in terms of best practices such as this.
I recommend to use Ruff or some other linter of choice and run it over the codebase.

PR #195, opened 2024-08-07
Schedule synthetic jobs relative to epoch

agoncharov-reef:
This PR evenly distributes validators' synthetic jobs across epoch blocks.
1) `schedule_synthetic_jobs` is run every minute, but it's supposed to really run at the beginning of an epoch and schedule a validation run by creating `ScheduledSyntheticJobs` object.
2) `run_synthetic_jobs` is run every minute and checks whether there is some unprocessed `ScheduledSyntheticJobs` instance. In case of such, spawn a `_run_synthetic_jobs` task.


validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -66,6 +67,97 @@ class WeightsRevealError(Exception):
...
+            this_validator_index = [vali.hotkey for vali in validators].index(this_hotkey)
+        except ValueError:
+            logger.exception(
+                "This validator is not in a list of validators -> not scheduling validation run"
+            )
+            return
+
+        next_run_block = when_to_run(
+            start_block=current_epoch[0],
+            end_block=current_epoch[1],
```
andreea-popescu-reef, line 152: given that get_epoch_containing_block is a range aren't these just consecutive blocks? shouldn't get_epoch_containing_block just output (epoch_start, epoch_end + 1) tuple?

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -80,28 +172,44 @@ def _run_synthetic_jobs():
 
 
 @app.task()
-def run_synthetic_jobs():
+def run_synthetic_jobs(max_late_blocks: int = 3) -> None:
```
andreea-popescu-reef, line 175: shouldn't max_late_blocks be bigger than the run_synthetic_jobs interval (which at every minutes is 5 blocks) or we might miss a ScheduledSyntheticJobs?

validator/app/src/compute_horde_validator/validator/tasks.py
agoncharov-reef: > just output (epoch_start, epoch_end + 1) tuple
Yeah, that's what ranges are exactly for :P

validator/app/src/compute_horde_validator/validator/tasks.py
agoncharov-reef: Fair point! Thank you!

validator/app/src/compute_horde_validator/validator/tests/conftest.py
```diff
@@ -53,3 +53,8 @@ def side_effect(*args, **kwargs):
 
         MockedMinerClient.side_effect = side_effect
         yield MockedMinerClient
+
+
+@pytest.fixture
+def validators():
+    return [MockNeuron(hotkey=f"hehetrololo{i}", uid=i) for i in range(10)]
```
ppolewicz, line 60:
```suggestion
    return [MockNeuron(hotkey=f"mock_validator_hotkey_{i}", uid=i) for i in range(10)]
```

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -66,6 +67,97 @@ class WeightsRevealError(Exception):
...
+
+@app.task
+def schedule_synthetic_jobs() -> None:
+    """
+    For current epoch, decide when miners' validation should happen.
+    Result is a ScheduledValidationRun object in the database.
+    """
+    with save_event_on_error(SystemEvent.EventSubType.GENERIC_ERROR), transaction.atomic():
+        try:
+            get_advisory_lock(LockType.VALIDATION_SCHEDULING)
```
ppolewicz, line 114: where is it released? Shouldn't this be a context manager?

validator/app/src/compute_horde_validator/validator/locks.py
```diff
@@ -3,20 +3,21 @@
...
     Has to be executed in transaction.atomic context. Throws `Locked` if not able to obtain the lock. The lock
     will be released automatically after transaction.atomic ends.
     """
     cursor = connection.cursor()
-    cursor.execute("SELECT pg_try_advisory_lock(%s)", [LockType.WEIGHT_SETTING])
+    cursor.execute("SELECT pg_try_advisory_lock(%s)", [type_])
     unlocked = cursor.fetchall()[0][0]
     if not unlocked:
-        raise Locked()
+        raise Locked
```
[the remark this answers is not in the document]
agoncharov-reef: This is a better stuff, Krszystof ported it from some library with few additions:
https://github.com/reef-technologies/django-subscriptions-rt/blob/master/subscriptions/utils.py#L69

validator/app/src/compute_horde_validator/validator/tasks.py
agoncharov-reef: It's run inside a transaction

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -66,11 +67,101 @@ class WeightsRevealError(Exception):
...
+    return epoch.start + floor(blocks_between_runs * (index_ + 1))
+
+
+def get_epoch_containing_block(block: int, netuid: int, tempo: int = 360) -> range:
+    """
+    Reimplementing the logic from subtensor's Rust function:
+        pub fn blocks_until_next_epoch(netuid: u16, tempo: u16, block_number: u64) -> u64
+    See https://github.com/opentensor/subtensor.
+
+    An epoch happens whenever (block_number + netuid + 1) is exactly divisible by (tempo + 1).
```
andreea-popescu-reef, line 95: the formula is weird though - the epoch block drift by 1 after each tempo ...?

agoncharov-reef: Yeah, i understand shit about it

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -79,29 +170,51 @@ def _run_synthetic_jobs():
...
-            raise ValueError(
-                "Can't determine proper synthetic job window due to not in top 23 validators"
+    subtensor_ = get_subtensor(network=settings.BITTENSOR_NETWORK)
+    current_block = subtensor_.get_current_block()
+
+    with transaction.atomic():
+        scheduled_run = (
+            ScheduledSyntheticJobs.objects.select_for_update(skip_locked=True)
+            .filter(
+                block__gte=current_block - max_late_blocks,
```
mzukowski-reef, line 201: I am concerned about this short window - schedule is 30s and so 3 blocks - and getting subtensor can take some time... I feel that we try to be either too exact or we don't try enough - I mean - if we really care about to schedule at right moment - we should check if current block is "near" schedule block and then iterate waiting for passing schedule block - also if for some reason (eg. restart) we are late like 1-3min we should still (it is my opinion - I might be wrong :D) schedule synthetic jobs

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -79,29 +170,51 @@ def _run_synthetic_jobs():
...
+                **(
+                    dict(block__lte=current_block)
+                    if not settings.DEBUG_DONT_STAGGER_VALIDATORS
```
mzukowski-reef, line 204: I think this completly loose its purpose - I use it on testnet to schedule synthetic jobs every 5min - I change run_synthetic_jobs schedule and set this and it works - now this is just useless - it would be better if setting DEBUG_DONT_STAGGER_VALIDATORS would completly ignore syncing synthethic jobs with epoch - still settings setting weights should be synced with epoch (right now on testnet it is set to schodule every 1h - after adding weight setting epoch sync it would actually not require hack on testnet)

validator/app/src/compute_horde_validator/validator/tasks.py
mpnowacki-reef: 3 minutes late is no no because then we will collide with another validator. the job takes ~45s to complete but they have 4minutes to set everything up

validator/app/src/compute_horde_validator/validator/locks.py
```diff
@@ -3,20 +3,21 @@
 
 class LockType:
```
bswck, line 4: Shouldn't that be an enum? (To be addressed in a separate PR, I'm asking due to it being used as a type annotation which feels inherently wrong)

bswck (commented):
One nit, not crucial

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -79,29 +170,51 @@ def _run_synthetic_jobs():
...
-        validator_keys = [n.hotkey for n in validators]
-        if my_key not in validator_keys:
-            raise ValueError(
-                "Can't determine proper synthetic job window due to not in top 23 validators"
+    subtensor_ = get_subtensor(network=settings.BITTENSOR_NETWORK)
+    current_block = subtensor_.get_current_block()
+
+    with transaction.atomic():
+        scheduled_run = (
+            ScheduledSyntheticJobs.objects.select_for_update(skip_locked=True)
```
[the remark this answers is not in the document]
agoncharov-reef: We shouldn't have more than one object by design, but in case there are multiple - we should run once only:
https://github.com/backend-developers-ltd/ComputeHorde/commit/603e4b8f37527301cbad44e5054fd675426d2027

validator/app/src/compute_horde_validator/validator/tasks.py
agoncharov-reef: @mpnowacki-reef Current approach gives us max 30s delay while we may do what @mzukowski-reef suggests and start waiting for right moment in advance. Do we care?

validator/app/src/compute_horde_validator/validator/locks.py
agoncharov-reef: Ah yeah, my type hint sucks :(

validator/app/src/compute_horde_validator/validator/tasks.py
agoncharov-reef: Ah, already discussed on slack

mpnowacki-reef: btw we definitely need a system event for "I overslept"

mpnowacki-reef: and the 30s should be a dynamic config param

mpnowacki-reef: oversleeping to dump "my block to start" is covered, right?

validator/app/src/compute_horde_validator/settings.py
```diff
@@ -139,6 +141,16 @@ def wrapped(*args, **kwargs):
         float,
     ),
     "DYNAMIC_WEIGHTS_VERSION": (1, "The weights version for synthetic jobs", int),
+    "SYNTHETIC_JOBS_PLANNER_WAIT_IN_ADVANCE_BLOCKS": (
```
mzukowski-reef, line 144: I think Michał said it should be dynamic config - all dynamic config variables use prefix `DYNAMIC_` it is used by syncing logic

so almost all constance variables should be dynamic - only those controlled by validator operator are not dynamic

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -79,29 +170,51 @@ def _run_synthetic_jobs():
...
-        logger.info(
-            f"Sleeping for {my_window_starts_at:02f}s because I am {my_index} out of {len(validator_keys)}"
+            .first()
         )
-        time.sleep(my_window_starts_at)
+        if not scheduled_run:
+            logger.debug("No scheduled validation run found")
+            return
+
+        scheduled_run.started_at = now()
```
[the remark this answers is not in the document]
mzukowski-reef: this comment was not addressed - if you are addressing it this is just kind reminder (and you don't need to change its name or change logic to batch - I just would like it to have more data about epoch so we can align weight setting - also I thought that we included some offset at the beginning  - do we need offset at the end - for weight settting - so last validator have time to send jobs, receive results before synced weight setting should happen?

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -80,28 +171,78 @@ def _run_synthetic_jobs():
...
+    poll_interval = poll_interval or timedelta(seconds=config.SYNTHETIC_JOBS_PLANNER_POLL_INTERVAL)
+
+    subtensor_ = get_subtensor(network=settings.BITTENSOR_NETWORK)
+    current_block = subtensor_.get_current_block()
+
+    with transaction.atomic():
+        ongoing_scheduled_jobs = list(
+            ScheduledSyntheticJobs.objects.select_for_update(skip_locked=True)
+            .filter(
+                block__gte=current_block,
```
mzukowski-reef, line 205: Michal asked about missed window event - we should somehow check for blocks that block__lt=current_block, started_at__isnull=True, sent one event and flagged as missed

validator/app/src/compute_horde_validator/validator/tasks.py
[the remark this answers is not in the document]
agoncharov-reef: Yes, I addressed it, thank you, just haven't replied :[

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -403,15 +621,31 @@ def set_scores():
             hotkey_to_uid = {n.hotkey: n.uid for n in neurons}
             score_per_uid = {}
             batches = list(
-                SyntheticJobBatch.objects.prefetch_related("synthetic_jobs").filter(
+                SyntheticJobBatch.objects.prefetch_related("synthetic_jobs")
+                .filter(
                     scored=False,
                     started_at__gte=now() - timedelta(days=1),
                     accepting_results_until__lt=now(),
+                    cycle__isnull=False,
```
mzukowski-reef, line 629: settings.DEBUG_DONT_STAGGER_VALIDATORS creates "empty" batch - testnet vali is using this to schedule jobs every 5min - then it set weights every 1h - not sure if commit reveal is enabled on testnet but this will break testnet validator weights setting

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -403,15 +621,31 @@ def set_scores():
...
                 )
+                .order_by("-started_at")
             )
             if not batches:
                 logger.info("No batches - nothing to score")
                 return
+            if len(batches) > 1:
```
mzukowski-reef, line 636: on testnet we run multiple batches in 1h weights setting window - this would also break weight setting :)

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -403,15 +621,31 @@ def set_scores():
...
+                logger.error("Unexpected number batches eligible for scoring: %s", len(batches))
+                for batch in batches[:-1]:
+                    batch.scored = True
+                    batch.save()
+                batches = [batches[-1]]
+
+            if subtensor.get_current_block() <= batches[-1].cycle.stop:
```
mzukowski-reef, line 643: <3

(but breaking testnet weight setting - I am not saying that we have to fix it - we can break testnet weight setting and fix it later)

validator/app/src/compute_horde_validator/validator/tests/helpers.py
```diff
@@ -312,6 +345,10 @@ def check_system_events(
     )
 
 
-def always_same_result(result):
-    """Return a function which accepts everything and returns the same result"""
-    return lambda *args, **kwargs: result
+def patch_constance(config_overlay: dict):
```
mzukowski-reef, line 348: Enam claims that @pytest.mark.overrride_config works :D

mzukowski-reef (approved):

PR #197, opened 2024-08-09
Bash improvements

bswck:
- Some scripts used `pipefail` without `-o`, which doesn't make sense, so that's pretty much a bugfix
- Many scripts had `-eux -o` that were merged to `-euxo`
- I don't really like mixing `set` with shebang-located bash flags; for cohesion, I moved all shebang-placed flags into `set` commands

I also addressed one shellcheck issue in 68e3234.

Aside from these changes, I'd recommend:
- using `-E` for scripts that use subshells (there's a lot of them)
- using `command`/`builtin` where it makes sense
- copying env vars to scripts using `!#/usr/bin/env bash` instead of `!#/bin/bash` where that's not breaking
- not using `.sh` extension for Bash/other shell scripts, it's misleading

let me know if you wish some of these implemented :)

### DIFF
--- a/executor/bin/prepare-os.sh
+++ b/executor/bin/prepare-os.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 # Copyright 2020, Reef Technologies (reef.pl), All rights reserved.
-set -eux pipefail
+set -euxo pipefail
 
 DOCKER_BIN="$(command -v docker || true)"
 DOCKER_COMPOSE_BIN="$(command -v docker-compose || true)"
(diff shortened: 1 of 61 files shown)


executor/bin/prepare-os.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
```
bswck, line 1: How on earth is that Bourne shell with pipefail below...

```suggestion
#!/bin/bash
```

bswck: Ok, there's a lot of it

bswck:
E2E tests don't run -- looks like the shebang was correct, but the flags were unnecessary.

bswck:
Since we're getting some shellcheck errors here that I intentionally omitted, I'll just correct that now and see if workflows are passing.

PR #206, opened 2024-08-20
Remove old synthetic jobs flow

emnoor-reef:


validator/app/src/compute_horde_validator/validator/tests/test_utils.py
mzukowski-reef, line 1: noooooooooo....... :D

without it we don't have any unit tests - I think I've made them kind of generic so it should not be that hard to adjust to new flow - it is kind of disturbing that new flow don't have unit tests

mzukowski-reef (changes requested):

validator/app/src/compute_horde_validator/validator/tests/test_utils.py
emnoor-reef: As I initially suspected, the tests took waaaaaay long to adjust to the new flow. :'(

There are some minor TODOs left in there. I'll look at them later (too exhausted now) :D

mzukowski-reef (approved):

PR #223, opened 2024-08-29
Add django-less miner client in library

emnoor-reef:
- Move miner client for running organic jobs from validator to the library, removing django specific things.
- Add django specific things (system event models etc.) in the subclass in validator.

### DIFF
--- a/compute_horde/compute_horde/miner_client/organic.py
+++ b/compute_horde/compute_horde/miner_client/organic.py
@@ -1,0 +1,241 @@
+import asyncio
+import datetime
+import logging
+import time
+from functools import cached_property
+
+import bittensor
+
+from compute_horde.base_requests import BaseRequest
+from compute_horde.executor_class import ExecutorClass
+from compute_horde.miner_client.base import AbstractMinerClient, UnsupportedMessageReceived
+from compute_horde.mv_protocol import miner_requests, validator_requests
+from compute_horde.mv_protocol.miner_requests import (
+    BaseMinerRequest,
+    UnauthorizedError,
+    V0AcceptJobRequest,
+    V0DeclineJobRequest,
+    V0ExecutorFailedRequest,
+    V0ExecutorManifestRequest,
+    V0ExecutorReadyRequest,
+    V0JobFailedRequest,
+    V0JobFinishedRequest,
+    V0MachineSpecsRequest,
+)
+from compute_horde.mv_protocol.validator_requests import (
+    AuthenticationPayload,
+    JobFinishedReceiptPayload,
+    JobStartedReceiptPayload,
+    V0AuthenticateRequest,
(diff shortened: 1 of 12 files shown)


compute_horde/compute_horde/miner_client/organic.py
```diff
@@ -0,0 +1,426 @@
...
+    raw_script: str | None = None
+    docker_run_options_preset: Literal["nvidia_all", "none"] = "nvidia_all"
+    docker_run_cmd: list[str] = field(default_factory=list)
+    total_job_timeout: int = 300
+    volume: Volume | None = None
+    output: OutputUpload | None = None
+
+    def __post_init__(self):
+        if self.base_docker_image_name is None:
+            self.base_docker_image_name = self.docker_image
```
olzhasar-reef, line 332: This is a bit confusing tbh, feels like `docker_image` could be populated from the `base_docker_image`, but the other way around...

compute_horde/tests/test_run_organic_job.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self, name: str, messages: list[str], *args, **kwargs):
+        super().__init__(name, messages, *args, **kwargs)
+        self.sent_models = []
+
+    async def send(self, message):
+        await super().send(message)
+        self.sent_models.append(BaseValidatorRequest.parse(message))
+
+
+class MockOrganicMinerClient(OrganicMinerClient):
```
olzhasar-reef, line 34: We can consider injecting the miner to the function to avoid mocking like this

compute_horde/compute_horde/miner_client/organic.py
```diff
@@ -0,0 +1,426 @@
...
+
+        if (self.docker_image, self.raw_script) == (None, None):
+            raise ValueError("At least of of `docker_image` or `raw_script` must be not None")
+
+
+async def run_organic_job(
```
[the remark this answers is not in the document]
emnoor-reef: Mostly yes.

compute_horde/compute_horde/miner_client/organic.py
emnoor-reef: I was thinking about removing `base_docker_image_name` and only have `docker_image` here, as it is used in a helper function to reduce complexity of users. Now that you've said something similar, I'll go ahead and remove it. :D

compute_horde/compute_horde/miner_client/organic.py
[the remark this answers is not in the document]
emnoor-reef: We may also refactor the validator's organic job code with this helper later. I'll check it's viability after we are done with our priority tasks.

olzhasar-reef (approved):

PR #225, opened 2024-08-31
Multiple executor classes support

olzhasar-reef:


miner/app/src/compute_horde_miner/miner/miner_consumer/validator_interface.py
```diff
@@ -277,7 +277,9 @@ async def handle(self, msg: BaseValidatorRequest):
             self.pending_jobs[msg.job_uuid] = job
 
             try:
-                await current.executor_manager.reserve_executor(token)
```
olzhasar-reef, line 280: This change is breaking. There is no clean way of soft deprecating this, cause the old method could possibly have custom logic implemented and handling both cases will arguably lead to a lot of obscurity.
The corresponding announcement about the breaking change has been in the discord channel by @mpnowacki-reef

mpnowacki-reef (approved):

PR #227, opened 2024-09-03
Prompt models

olzhasar-reef:
This PR contains:
- Prompt db models
- Configuration and utils for working with S3 in the validator


emnoor-reef (approved):

validator/app/src/compute_horde_validator/settings.py
```diff
@@ -465,6 +465,11 @@ def BITTENSOR_WALLET() -> bittensor.wallet:
     },
 }
 
+AWS_ACCESS_KEY_ID = env.str("AWS_ACCESS_KEY_ID")
+AWS_SECRET_ACCESS_KEY = env.str("AWS_SECRET_ACCESS_KEY")
+AWS_ENDPOINT_URL = env("AWS_ENDPOINT_URL", default=None)
+AWS_S3_BUCKET_NAME = env.str("AWS_S3_BUCKET_NAME")
```
emnoor-reef, line 471: Making these required is a breaking change. We need to  make sure validators add the env vars before deploying.

validator/app/src/compute_horde_validator/validator/models.py
```diff
@@ -296,3 +297,38 @@ def save(self, *args, **kwargs) -> None:
...
+class PromptBatch(models.Model):
+    batch_uuid = models.UUIDField(default=uuid.uuid4, unique=True)
+    s3_url = models.URLField(max_length=1000)
+    created_at = models.DateTimeField(default=now)
+    generator_version = models.PositiveSmallIntegerField(default=settings.PROMPT_GENERATOR_VERSION)
+
+
+class SampleBatch(models.Model):
+    batch_uuid = models.UUIDField(default=uuid.uuid4, unique=True)
+    seed = models.IntegerField()
```
emnoor-reef, line 311: Should be a BigIntegerField (64-bit). IntegerField. is 32-bit AFAIK. Seed supports 128-bit integers, but we can stay in the 64-bit integer limits.

validator/app/src/compute_horde_validator/settings.py
olzhasar-reef: Great point. Made these optional, we can raise a warning instead while skipping the entire flow later

emnoor-reef (approved):

PR #230, opened 2024-09-05
Sample

andreea-popescu-reef:


validator/app/src/compute_horde_validator/settings.py
```diff
@@ -437,6 +437,15 @@ def wrapped(*args, **kwargs):
 
 DYNAMIC_CONFIG_ENV = env.str("DYNAMIC_CONFIG_ENV", default="prod")
 
+# prompt gen sampling
```
mpnowacki-reef, line 440: these have to be dynamic params, not read from env vars

validator/app/src/compute_horde_validator/validator/models.py
```diff
@@ -315,6 +315,9 @@ class SampleBatch(models.Model):
     s3_url = models.URLField(max_length=1000)
```
mpnowacki-reef, line 315: @olzhasar-reef @andreea-popescu-reef @emnoor-reef Should SampleBatch point to a single SyntheticJobBatch? is it even needed? If a SampleBatch produces validation data for 120 prompt batches, these batches can easily be used across various synthetic job batches.

mpnowacki-reef: maybe that alright, some will go to waste

mpnowacki-reef: though in the beginning it's gonna be most of them

mpnowacki-reef: but maybe that's alright too, we'll test how validator work on generating new ones, this will help us simulate a case where they actually use up all the stuff

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -952,3 +956,69 @@ def fetch_dynamic_config() -> None:
...
+
+        # sample lines
+        sampled_lines = random.sample(
+            lines, min(settings.DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_IN_BATCH, len(lines))
+        )
+
+        # save the sampled prompts as unanswered in the db
+        Prompt.objects.bulk_create(
+            [Prompt(prompt_batch=prompt_batch, content=line) for line in sampled_lines]
+        )
```
emnoor-reef, line 981: If we create `Prompt` for all the `PromptBatch`, and then later make a `SampleBatch` with `batch_size` prompts, it can contain a lot of prompts from some `PromptBatch`s, and very few from others.

Instead of `DYNAMIC_NUMBER_OF_PROMPTS_IN_BATCH`, we can have `DYNAMIC_NUMBER_OF_BATCHES_PER_SAMPLE`, so that:

number of prompts in sample batch = `DYNAMIC_NUMBER_OF_BATCHES_PER_SAMPLE * DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_IN_BATCH`

This flow should simplify it a bit:
- Get a chunks of `PromptBatch`s of size `DYNAMIC_NUMBER_OF_BATCHES_PER_SAMPLE`
- For each chunk:
  - create a `SampleBatch` with a seed and a boolean `has_completed` or something saying the sampling has not completed.
  - create `DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_IN_BATCH` `Prompt` for each `PromptBatch` and save the `sample_batch` with the one created above.
  - trigger the task to do the job to get the answers (this task should update `has_completed` field of the `SampleBatch` after updating the `Prompt.answer`s)

This would ensure each each `SampleBatch` would have the same number of `Prompt`s from each `PromptBatch`s. It would also simplify a bit in the sending of synthetic jobs to miners.

emnoor-reef (commented):
Looks like Olzhas's push broke your branch, needs a rebase.

I'm only reviewing the file `validator/app/src/compute_horde_validator/validator/tasks.py`.

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -952,3 +962,58 @@ def fetch_dynamic_config() -> None:
...
+        # should always have enough prompts
+        if len(lines) <= prompts_per_sample:
+            logger.error("Skipping bucket %s, not enough prompts", prompt_series.s3_url)
+            continue
+
+        # sample prompts
+        sampled_lines = random.sample(lines, prompts_per_sample)
+        current_workload_fill += len(sampled_lines)
+
+        prompt_sample = PromptSample.objects.create(series=prompt_series, workload=current_workload)
```
olzhasar-reef, line 1014: As all these queries are related and does not make much sense independently, it's probably a good idea to use a db transaction here

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -952,3 +962,58 @@ def fetch_dynamic_config() -> None:
         config_url=f"https://raw.githubusercontent.com/backend-developers-ltd/compute-horde-dynamic-config/master/validator-config-{settings.DYNAMIC_CONFIG_ENV}.json",
         namespace=config,
     )
+
+
+def create_workload(seed: int):
+    # generate an s3 url to upload sample batch job result in
+    workload_uuid = uuid.uuid4()
+    s3_url = generate_upload_url(key=workload_uuid, bucket_name="upload-bucket-{sample_batch.uuid}")
```
olzhasar-reef, line 970: I'm not sure why is the bucket name dynamic here
Btw, I've added two settings for s3 buckets:
```
S3_BUCKET_NAME_PROMPTS
S3_BUCKET_NAME_ANSWERS
```
Here, we should be using the second one (ANSWERS)

validator/app/src/compute_horde_validator/validator/dynamic_config.py
```diff
@@ -31,6 +31,24 @@ async def aget_config(key):
...
+async def aget_number_of_prompts_to_validate_from_series():
+    if settings.DEBUG_OVERRIDE_DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_FROM_SERIES is not None:
+        return settings.DEBUG_OVERRIDE_DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_FROM_SERIES
+    return await aget_config("DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_FROM_SERIES")
+
+
+async def aget_number_of_workloads_to_trigger_local_inference():
+    if settings.DEBUG_OVERRIDE_DYNAMIC_NUMBER_OF_WORKLOADS_TO_TRIGGER_LOCAL_INFERENCE is not None:
+        return settings.DEBUG_OVERRIDE_DYNAMIC_NUMBER_OF_WORKLOADS_TO_TRIGGER_LOCAL_INFERENCE
+    return await aget_config("DYNAMIC_NUMBER_OF_WORKLOADS_TO_TRIGGER_LOCAL_INFERENCE")
```
emnoor-reef, line 49: These settings needs to be in `settings.CONSTANCE_CONFIG`.

validator/app/src/compute_horde_validator/validator/dynamic_config.py
```diff
@@ -31,6 +31,24 @@ async def aget_config(key):
...
+async def aget_number_of_prompts_in_series():
+    if settings.DEBUG_OVERRIDE_DYNAMIC_NUMBER_OF_PROMPTS_IN_SERIES is not None:
+        return settings.DEBUG_OVERRIDE_DYNAMIC_NUMBER_OF_PROMPTS_IN_SERIES
+    return await aget_config("DYNAMIC_NUMBER_OF_PROMPTS_IN_SERIES")
+
+
+async def aget_number_of_prompts_to_validate_from_series():
+    if settings.DEBUG_OVERRIDE_DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_FROM_SERIES is not None:
+        return settings.DEBUG_OVERRIDE_DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_FROM_SERIES
+    return await aget_config("DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_FROM_SERIES")
```
emnoor-reef, line 43: These functions seem to be only used in sync context and not in a hot-loop. So sync-to-async-to-sync conversion seems overkill. (same for the other two)

```suggestion
def get_number_of_prompts_to_validate_from_series():
    if settings.DEBUG_OVERRIDE_DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_FROM_SERIES is not None:
        return settings.DEBUG_OVERRIDE_DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_FROM_SERIES
    return config.DYNAMIC_NUMBER_OF_PROMPTS_TO_VALIDATE_FROM_SERIES
```

emnoor-reef: If `DEBUG_` settings are not strictly needed (as you can easily override these in tests), I'd remove these functions altogether and use `config.` directly at the call-sites.

PR #232, opened 2024-09-06
Coinstar test task

lemeote:


ppolewicz:
(it doesn't work, is generally unfinished with the hardvoded v1.0.0 etc)

PR #236, opened 2024-09-11
revert miner bad rebase

andreea-popescu-reef:


olzhasar-reef:
Don't you want to save this tests though?

mzukowski-reef:
> Don't you want to save these tests though?

it is some work in progress test file :D 

and it is already in history - you can always recover it :)

mzukowski-reef (approved):

PR #240, opened 2024-09-12
llm prompts app tasks

andreea-popescu-reef:


validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1134,3 +1143,75 @@ def create_sample_workloads():
...
+
+async def get_workload_prompts(workload: SolveWorkload) -> list[Prompt]:
+    sample_prompts = PromptSample.objects.prefetch_related("prompts").filter(
+        workload=workload,
+        prompts__answer__isnull=True,
+    )
+    prompts = []
+    async for sample in sample_prompts:
+        prompts.extend(sample.prompts.all())
+    return prompts
```
olzhasar-reef, line 1156: Why not:
```
return Prompt.objects.select_related('sample').filter(sample__workload_id=workload.id, answer__isnull=True)
```
? 😁

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1134,3 +1143,75 @@ def create_sample_workloads():
...
+        except Exception:
+            logger.error("Failed to download prompt answers", exc_info=True)
+            continue
+
+        # update the workload as finished
+        workload.finished_at = now()
+        await workload.asave()
+
+        # update the prompts with the answers
+        for prompt in prompts:
```
olzhasar-reef, line 1212: If the server shuts down and the execution stops somewhere here, we will end up having inconsistent data.
We should consider having a separate sync function wrapped in an atomic block and call it with sync_to_async or smth

validator/app/src/compute_horde_validator/validator/tests/settings.py
```diff
@@ -45,6 +45,6 @@ def BITTENSOR_WALLET() -> bittensor.wallet:
 
 DYNAMIC_CONFIG_CACHE_TIMEOUT = 0
 
-GENERATION_MINER_KEY = "fake_generation_miner_key"
-GENERATION_MINER_ADDRESS = "fakehost"
-GENERATION_MINER_PORT = 1234
+TRUSTED_MINER_KEY = "fake_generation_miner_key"
```
olzhasar-reef, line 48: Will we have just a single trusted miner for now?

validator/app/src/compute_horde_validator/validator/s3.py
```diff
@@ -51,3 +52,10 @@ def get_prompts_from_s3_url(s3_url: str) -> Generator[tuple[str, list[str]]]:
         logger.warning(f"Failed to download prompts from {s3_url}")
         return []
     return response.text.split("\n")
+
+
+async def download_json(s3_url: str):
```
olzhasar-reef, line 57: I would call it just `download_file_content` or smth, cause the function doesn't necessarily assume json inside

validator/app/src/compute_horde_validator/validator/tests/settings.py
andreea-popescu-reef: I thought that was the initial assumption - but can separate the vars anyway

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -762,10 +762,7 @@ async def _close_client(ctx: BatchContext, miner_hotkey: str) -> None:
         await client.close()
 
 
-async def _generate_jobs(ctx: BatchContext) -> None:
-    start_time = time.time()
-    generated_job_count = 0
-
+async def get_llama_prompt_samples(ctx: BatchContext) -> PromptSample | None:
```
emnoor-reef, line 765:
```suggestion
async def get_llama_prompt_samples(ctx: BatchContext) -> list[PromptSample] | None:
```

validator/app/src/compute_horde_validator/validator/synthetic_jobs/generator/llama_prompts.py
```diff
@@ -1,30 +1,32 @@
...
+        prompt_sample: PromptSample | None,
         expected_prompts: list[Prompt],
         s3_url: str,
         seed: int,
         **kwargs,
     ):
         super().__init__(**kwargs)
-
-        self.prompt_sample: PromptSample = prompt_sample
+        self.prompt_sample: PromptSample | None = prompt_sample
```
emnoor-reef, line 29: `self.prompt_sample` cannot be `None`. If there's no PromptSample, this generator should never be called in the first place. If it is None, this will fail in the synthetic jobs flow: https://github.com/backend-developers-ltd/ComputeHorde/blob/7c20ce09534112016ada8a0d3a4221568c818e28/validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py#L1419-L1426

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -781,8 +778,24 @@ async def _generate_jobs(ctx: BatchContext) -> None:
...
+        )
+        return None
+    return llama_prompt_samples
+
+
+async def _generate_jobs(ctx: BatchContext) -> None:
+    start_time = time.time()
+    generated_job_count = 0
+
+    llama_prompt_samples = await get_llama_prompt_samples(ctx)
```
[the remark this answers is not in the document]
andreea-popescu-reef: no - we only send llama to llama tasks executors https://github.com/backend-developers-ltd/ComputeHorde/pull/240/files/0b24ef06466029ae6530b3a379b3dcb086b56806#diff-399b6cde1b3f4939e5344ce72773b39b38e23cad3c26010f1b75736c6ff3ca31R807

validator/app/src/compute_horde_validator/validator/synthetic_jobs/generator/llama_prompts.py
andreea-popescu-reef: so there are 2 flows - one for the synthetic jobs where indeed it can't be None - and one for generating the answers for the workload (the workload contains multiple prompts associated with different prompt samples) where the prompt sample is not needed

andreea-popescu-reef: I'll try to add clearer check for it

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1082,9 +1080,50 @@ def create_workload(seed: int):
...
+        )
+
+
+# prompt_sample is ready for synthetic job when associated workload is finished (all prompts are answered)
+@app.task()
+def llama_prompt_answering():
+    # prioritize answering all the workloads before generating new samples
+    unprocessed_workloads = SolveWorkload.objects.filter(finished_at__isnull=True)
+    for workload in unprocessed_workloads:
+        async_to_sync(answer_prompts)(workload)
```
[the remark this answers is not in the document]
andreea-popescu-reef: good point - I'll handle it

validator/app/src/compute_horde_validator/settings.py
```diff
@@ -392,6 +400,16 @@ def wrapped(*args, **kwargs):
...
         "options": {},
     },
+    "llama_prompt_generation": {
+        "task": "compute_horde_validator.validator.tasks.llama_prompt_generation",
+        "schedule": timedelta(minutes=10),
+        "options": {},
+    },
+    "llama_prompt_answering": {
+        "task": "compute_horde_validator.validator.tasks.llama_prompt_answering",
+        "schedule": timedelta(minutes=10),
```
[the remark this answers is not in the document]
andreea-popescu-reef: yeah - I'm trying to run tests with smaller values to check for errors - the actual env vars values (not just the scheduling, but the timeouts for jobs too) will need to be figured out when we deploy

validator/app/src/compute_horde_validator/validator/cross_validation/prompt_answering.py
```diff
@@ -0,0 +1,109 @@
...
+    wait_timeout: int | None = None,
+) -> None:
+    if not all(
+        [
+            settings.TRUSTED_MINER_KEY,
+            settings.TRUSTED_MINER_ADDRESS,
+            settings.TRUSTED_MINER_PORT,
+        ]
+    ):
+        logger.warning("Prompt generation miner not configured, skipping prompt generation")
```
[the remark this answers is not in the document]
andreea-popescu-reef: it should be trusted miner - wouldn't we just add executors instead of additional miners if needed though?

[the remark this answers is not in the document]
olzhasar-reef: The thing is, I don't think we are currently able to properly route jobs for multiple executor classes out of the box. We added a general ability for that, but there needs to be a subclassed `DockerExecutorManager` with overwritten `get_manifest` method for this to work.
So I assume the idea for the upcoming release was to have different miners for now. but I am not sure if a single one is okay though, maybe it is. Let's discuss this tomorrow

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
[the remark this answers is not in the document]
olzhasar-reef: Oh, I see. Maybe we can skip the query altogether then if we don't have such executors? Either here or inside the get_llama_prompt_sample func.
But that's a minor thing though, can be ignored for now

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1031,63 +1034,132 @@ def fetch_dynamic_config() -> None:
...
 
-def create_workload(seed: int):
-    # generate an s3 url to upload sample batch job result in
+@app.task()
+def llm_prompt_generation():
+    num_expected_prompt_series = config.DYNAMIC_MAX_PROMPT_SERIES
+    num_prompt_series = PromptSeries.objects.count()
+    if num_prompt_series < num_expected_prompt_series:
+        logger.info("There are %s series in the db, generating prompts", num_prompt_series)
+        async_to_sync(generate_prompts)()
```
mzukowski-reef, line 1043: we need to lock local miner here - only one task can use local miner at the same time

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1031,63 +1034,132 @@ def fetch_dynamic_config() -> None:
...
+            "There are %s series in the db - skipping prompt generation",
+            num_prompt_series,
+        )
+
+
+@app.task()
+def llm_prompt_answering():
+    with transaction.atomic():
+        try:
+            get_advisory_lock(LockType.ANSWERING_PROMPTS)
```
mzukowski-reef, line 1055: don't know if we want to lock for all workloads - maybe we need to lock per workload allowing switching between prompt answering and generating

mzukowski-reef (changes requested):

miner/app/src/compute_horde_miner/miner/miner_consumer/validator_interface.py
```diff
@@ -367,7 +367,7 @@ async def handle(self, msg: BaseValidatorRequest):
                 time_took_us=msg.payload.time_took_us,
                 score_str=msg.payload.score_str,
             )
-            prepare_receipts.delay()
+            await prepare_receipts()
```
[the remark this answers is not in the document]
andreea-popescu-reef: ah I see - was triggering this from a local command without celery and was breaking  - will revert

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1031,63 +1034,132 @@ def fetch_dynamic_config() -> None:
...
+        PromptSample.objects.bulk_create(prompt_samples)
+        Prompt.objects.bulk_create(prompts)
+
+
+def create_sample_workloads(num_needed_prompt_samples):
+    prompts_per_sample = config.DYNAMIC_NUMBER_OF_PROMPTS_TO_SAMPLE_FROM_SERIES
+    prompts_per_workload = config.DYNAMIC_NUMBER_OF_PROMPTS_PER_WORKLOAD
 
     # set seed for the current synthetic jobs run
     seed = random.randint(0, 1000000)
```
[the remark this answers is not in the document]
andreea-popescu-reef: wait - that would not fit into the BigIntegerField
I guess we would be fine with 64bit?

validator/app/src/compute_horde_validator/validator/tasks.py
andreea-popescu-reef: we should always prioritize answering so it should be fine to no switch from this task unless it's finished - if there are unanswered workloads we don't gain anything by generating or sampling more

answering will run when there are workloads available - will not run otherwise
sampling will run when there are not enough workloads prepped
^ these can be triggered frequently as they will stop when there's nothing left to do
generation can run inbetween to refresh the db prompt series - I assume we'll be running only generation when deploying to fill the db and then switch on the other tasks

validator/app/src/compute_horde_validator/settings.py
```diff
@@ -403,6 +429,21 @@ def wrapped(*args, **kwargs):
         "schedule": timedelta(minutes=5),
         "options": {},
     },
+    # "llm_prompt_generation": {
```
mzukowski-reef, line 432: I kind don't get it how we plan it to interact - running them every 10min with 10min timeout seems kind of wrong - the idea was I think to have like a "router" task that would see what should be done and schedule a small batch of work - if this "router" would see that some work is being done it should finish task and let it check again in next run

current timeouts for tasks are I think wrong - we select everything in database (workload, series) and just run the flow for every object - this might not be aligned with timeout - we should take as much work as timeout allows (with some buffer) and run it

I think this is like design discussion - current implementation is ok for testing, but we need to decide for target solution

mzukowski-reef (changes requested):

validator/app/src/compute_horde_validator/settings.py
andreea-popescu-reef: the timeouts (and most of the new vars values) are not decided yet - we should at least have an idea of how long a generation/answering task takes before setting them. ~ 10m were testing placeholder.

yeah, I stuck with the idea of having separate tasks

to maximize the amount of work between synthetic jobs: we can just trigger generation -> sampling -> answering.  We can generate and sample exactly enough prompts/workloads for an answering run.

though I'm still confused with the generation tasks - I though we wanted to run it a lot in the beginning until we have enough prompt series and then use those / not continue generating than much

PR #241, opened 2024-09-12
Fix executor manager timeout handling

mzukowski-reef:


miner/app/src/compute_horde_miner/miner/executor_manager/_internal/base.py
```diff
@@ -71,26 +75,30 @@ async def _pool_cleanup_loop(self):
...
+            elif reserved_executor.is_expired():
+                await self.manager.kill_executor(reserved_executor.executor)
+                return reserved_executor, True
+            return reserved_executor, False
+
+        results = await asyncio.gather(
+            *[check_executor(reserved_executor) for reserved_executor in self._executors]
+        )
+
+        self._executors = [
```
olzhasar-reef, line 94:
```suggestion
        async with self._reservation_lock:
            self._executors = [
                reserved_executor for reserved_executor, should_drop in results if not should_drop
            ]
```
As we are modifying the same list

miner/app/src/compute_horde_miner/miner/tests/test_executor_manager.py
```diff
@@ -0,0 +1,210 @@
...
+    0,
+)
+@patch(
+    "compute_horde_miner.miner.executor_manager._internal.base.ExecutorClassPool.POOL_CLEANUP_PERIOD",
+    0.1,
+)
+async def test_manager_reserve_executor_class(dummy_manager):
+    await dummy_manager.reserve_executor_class("token1", ExecutorClass.always_on__gpu_24gb, 10)
+    assert (
+        dummy_manager._executor_class_pools[ExecutorClass.always_on__gpu_24gb].get_availability()
```
olzhasar-reef, line 154: I would to abstain from using implementation details like these in tests, otherwise it would be quite hard to refactor anything.
We can use a public interface for getting a specific executor pool here instead.

mzukowski-reef: public interface has side effect - so it is actually better to use private interface here - unless you actually want to test if this side effect works or do not mess anything

miner/app/src/compute_horde_miner/miner/executor_manager/_internal/base.py
mzukowski-reef: your suggestion is actually wrong - this would cause "deadlockish" situation - context having self._reservation_lock would wait until availablility gets > 0 but this context would wait until the other would timeout - this is asyncio so there would be no race condition on self._executors - if we would worry about thread safeness we would need another lock - but we don't - we know that only single thread would be accessing self._executors

olzhasar-reef: > this is asyncio so there would be no race condition on self._executors
This is not true, a race condition is very possible here.
Consider this line:
```
            results = await asyncio.gather(
                *[check_executor(reserved_executor) for reserved_executor in self._executors]
            )
```
We construct a list and then proceed to asynchronously process each of the inner tasks. At this point, event loop is free to switch to another coroutine. If it switches to `reserve_executor`, it will append a new entry to this list, which we don't account for in the previous `results` list. Then it switches back and we replace `self._executors` from the `results` ending up having inconsistent data.

You are partially correct about waiting for availability part. A simple solution is to introduce another lock: `_executors_lock` or smth. And use that whenever we are modifying the list.

miner/app/src/compute_horde_miner/miner/tests/test_executor_manager.py
olzhasar-reef: It's never "actually better" to use private interfaces in tests. Private interface are implementation details, and you don't want to test these. Otherwise, we won't have any room for internal changes without a burden of changing all the tests as well.
I don't understand which side effect are you referring to. There is a `get_manifest` call inside `get_executor_class_pool`, but we have full control over that, cause it's overwritten in the `DummyExecutorManager`

miner/app/src/compute_horde_miner/miner/executor_manager/_internal/base.py
mzukowski-reef: yeah... you are right :) completly forgot why I implemented it as 
```
        for reserved_executor in self._executors:
            if reserved_executor not in executors_to_drop:
                still_running_executors.append(reserved_executor)
 ```
 will revert to this logic again - but still locking here is wrong :)

File: miner/app/src/compute_horde_miner/miner/tests/test_executor_manager.py (hunk 2 above)
Comment (mzukowski-reef) [reply miner/app/src/compute_horde_miner/miner/tests/test_executor_manager.py:154]:
```
    async def get_executor_class_pool(self, executor_class):
        manifest = await self.get_manifest()
        for executor_class, executor_count in manifest.items():
            pool = self._executor_class_pools.get(executor_class)
            if pool is None:
                pool = ExecutorClassPool(self, executor_class, executor_count)
                self._executor_class_pools[executor_class] = pool
            else:
                pool.set_count(executor_count)
        return self._executor_class_pools[executor_class]
  ```
  wanted to avoid testing this - but if you insist I will comply :)

mzukowski-reef: @olzhasar-reef  tried it but your tests in validator fails now :D

olzhasar-reef (approved):

PR #245, opened 2024-09-16
Script for provisioning S3 buckets

olzhasar-reef:


validator/provision_s3.sh
```diff
@@ -0,0 +1,59 @@
...
+PROMPTS_BUCKET=$1
+ANSWERS_BUCKET=$2
+
+BUCKET_REGION="${AWS_REGION:-us-east-1}"
+
+create_bucket() {
+    local bucket_name=$1
+
+    if aws s3api head-bucket --bucket "$bucket_name" 2>/dev/null; then
+      echo "Bucket $bucket_name already exists. Skipping creation."
```
mzukowski-reef, line 23: we might want to raise error when bucket name is aleady used by someone else - Claude suggests:
```
    if aws s3api head-bucket --bucket "$bucket_name" 2>/dev/null; then
        echo "Bucket $bucket_name already exists in your account. Skipping creation."
        return
    elif [ $? -eq 254 ]; then
        echo "Error: Bucket $bucket_name exists but you don't have access to it."
        echo "This could mean the bucket is owned by another AWS account."
        exit 1
    fi
    ```
    didn't test it

mzukowski-reef (approved):

validator/provision_s3.sh
olzhasar-reef: Just tested that, it turns out aws cli will return 254 both when bucket does not exist at all and when it's taken by someone else. So I have just made it exit with a code 1 for now instead of skipping.

PR #247, opened 2024-09-19
Fix commit reveal

olzhasar-reef:


validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -644,20 +626,55 @@ def get_metagraph(subtensor, netuid):
...
+
+    if commit_reveal_weights_enabled:
+        commit_reveal_weights_interval = config.DYNAMIC_COMMIT_REVEAL_WEIGHTS_INTERVAL
+        commit_start_offset = config.DYNAMIC_COMMIT_REVEAL_COMMIT_START_OFFSET
+        commit_end_buffer = config.DYNAMIC_COMMIT_REVEAL_COMMIT_END_BUFFER
+        current_interval = _get_commit_reveal_interval_for_block(
+            current_block, commit_reveal_weights_interval
+        )
+
+        if (
```
mzukowski-reef, line 656:
```
commit_window_start = current_interval.start + commit_start_offset
commit_window_end = current_interval.stop - commit_end_buffer

if not (commit_window_start <= current_block < commit_window_end):

```
or maybe just put `not` outside bool expression - but my version seems more readable :)

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -644,20 +626,55 @@ def get_metagraph(subtensor, netuid):
...
+            not current_interval.start + commit_start_offset
+            <= current_block
+            < current_interval.stop - commit_end_buffer
+        ):
+            logger.debug(
+                "Outside of commit window, skipping, current block: %s, interval: %s, offset: %s, buffer: %s",
```
mzukowski-reef, line 662: Can we have "somewhere" a nice summary what is happenning here...?

I mean - how cycle start/end interacts with weights commiting and weights revealing, why we have start-offset... 

why we wait with commit till end of a cycle - do we have to?

what is the logic of computing emissions for given cycle - possible cases analyzed... 

it is really hard to review this in terms of logic if nothing is documented

mzukowski-reef (changes requested):

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -189,6 +190,85 @@ def get_cycle_containing_block(block: int, netuid: int, tempo: int = 360) -> ran
     return range(first_epoch.start, second_epoch.stop)
 
 
+class CommitRevealInterval:
```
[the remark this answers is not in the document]
olzhasar-reef: Reusing the logic for calculating cycles was the first thing I had in mind.
But the issue is that the commit reveal intervals calculation is not tied to subnets in the actual subtensor code.
And we could use a netuid of 0, but this would arguably make things more confusing.
Speaking of dynamic params, for now it's easier to share the common logic like that (cause we still use dynamic params everywhere), but I made everything overridable in the __init__ to have a bit of a room there.

mzukowski-reef (approved):

PR #252, opened 2024-09-25
Hashcat encryption

kkowalski-reef:
- Encrypts the first stage of the hashcat synthetic job
- Uses miner hotkey as part of the key for the first stage

### DIFF
--- a/validator/app/src/compute_horde_validator/validator/synthetic_jobs/v2_synthetic_job.py
+++ b/validator/app/src/compute_horde_validator/validator/synthetic_jobs/v2_synthetic_job.py
@@ -37,7 +37,11 @@ class V2SyntheticJob(SyntheticJob):
 
     @classmethod
     def generate(
-        cls, algorithms: list[Algorithm], params: list[JobParams], salt_length_bytes: int = 8
+        cls,
+        algorithms: list[Algorithm],
+        params: list[JobParams],
+        miner_hotkey: str,
+        salt_length_bytes: int = 8,
     ) -> Self:
         # generate distinct passwords for each algorithm
         passwords = []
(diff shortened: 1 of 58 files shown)


validator/app/src/compute_horde_validator/validator/synthetic_jobs/v2_synthetic_job.py
```diff
@@ -0,0 +1,142 @@
...
+            return file.read()
+
+    @property
+    def answer(self) -> str:
+        return self._hash(
+            "".join(["".join(passwords) for passwords in self.passwords]).encode("utf-8")
+        ).decode("utf-8")
+
+    def __str__(self) -> str:
+        return f"V2SyntheticJob {self.algorithms} {self.params}"
```
agoncharov-reef, line 133:
```suggestion
        return f"{self.__class__.__name__} {self.algorithms} {self.params}"
```

validator/app/src/compute_horde_validator/validator/synthetic_jobs/v2_synthetic_job.py
```diff
@@ -0,0 +1,142 @@
...
+            "num_digits": [params.num_digits for params in self.params],
+        }
+        return pickle.dumps(data)
+
+    def docker_run_cmd(self) -> list[str]:
+        return ["/script.py", self.first_password]
+
+    def raw_script(self) -> str:
+        with open(Path(__file__).parent / "v2_decrypt.py") as file:
+            return file.read()
```
agoncharov-reef, line 124:
```suggestion
        return (Path(__file__).parent / "v2_decrypt.py").read_text()
```

agoncharov-reef (approved):
I think copy-paste for new versions of models is totally OK. No dependencies is good.

validator/app/src/compute_horde_validator/validator/synthetic_jobs/v2_synthetic_job.py
```diff
@@ -0,0 +1,142 @@
...
+            _passwords = set()
+            while len(_passwords) < _params.num_hashes:
+                _passwords.add(
+                    cls.random_string(
+                        num_letters=_params.num_letters, num_digits=_params.num_digits
+                    )
+                )
+            passwords.append(sorted(list(_passwords)))
+
+        first_password = f"{miner_hotkey}-{cls.random_string(num_letters=48, num_digits=0)}"
```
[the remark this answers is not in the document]
mzukowski-reef: num_digits is only for hashcat - so we have better control over password space - meaning - breaking password with additional letter is much harder than with additional digit or two - this password is different - it is just a random string - 48 letters makes it unbreakable :)

PR #258, opened 2024-09-26
Initial job request volume

kkowalski-reef:
- Add optional `volume` field to initial job request without bumping protocol version
- Volume cannot be sent in initial job request and job request at the same time
- Move the volume in synthetic hashcat jobs to the initial job request

### DIFF
--- a/executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
+++ b/executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
@@ -19,8 +19,8 @@ from compute_horde.base.volume import (
     InlineVolume,
     MultiVolume,
     SingleFileVolume,
-    ZipUrlVolume,
     Volume,
+    ZipUrlVolume,
 )
 from compute_horde.base_requests import BaseRequest
 from compute_horde.em_protocol import executor_requests, miner_requests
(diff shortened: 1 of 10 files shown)


executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -598,10 +600,16 @@ async def _unpack_multi_volume(self, volume: MultiVolume):
             else:
                 raise NotImplementedError(f"Unsupported sub-volume type: {type(sub_volume)}")
 
-    async def unpack_volume(self, job_request: V0JobRequest):
+    async def get_job_volume(self) -> Volume | None:
+        if self.full_job_request.volume and self.initial_job_request.volume:
+            raise JobError("Received multiple volumes")
+
+        return self.full_job_request.volume or self.initial_job_request.volume or None
```
agoncharov-reef, line 607: Shouldn't be async imo

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -812,7 +812,7 @@ async def _send_initial_job_request(
         executor_class=job.executor_class,
         base_docker_image_name=job.job_generator.base_docker_image_name(),
         timeout_seconds=job.job_generator.timeout_seconds(),
-        volume_type=VolumeType.inline,
+        volume=InlineVolume(contents=job.volume_contents),
```
agoncharov-reef, line 815: I may be wrong (since my knowledge is outdated and I don't see bigger picture), but as far as I understand the idea was to 1) send initial job request to allow miner prepare image 2) send actual payload when miner is ready to accept jobs.
With this change actual payload is sent immediately. Is this intended?

agoncharov-reef (commented):
In order to not trigger re-review of already reviewed stuff, you should've set `hashcat-encryption` as target branch, thus I would review only changes in `initial-job-request-volume` branch instead of changes in both branches.

mzukowski-reef (approved):

PR #261, opened 2024-10-01
Fix batch scoring N+1 query

olzhasar-reef:


validator/app/src/compute_horde_validator/validator/scoring.py
```diff
@@ -63,7 +63,7 @@ def score_jobs(jobs, score_aggregation=sum, normalization_weight=1):
 def score_batch(batch):
     executor_class_weights = get_executor_class_weights()
     executor_class_jobs = defaultdict(list)
-    for job in batch.synthetic_jobs.all():
+    for job in batch.synthetic_jobs.select_related("miner"):
```
mzukowski-reef, line 66: I think this is wrong - synthethic_jobs are prefetched - so we need to select_related on prefetched query :)

mzukowski-reef (changes requested):
I

validator/app/src/compute_horde_validator/validator/scoring.py
mzukowski-reef: well - actually I am wrong :D

mzukowski-reef (approved):

validator/app/src/compute_horde_validator/validator/scoring.py
olzhasar-reef: Well, your point is generally legit, however having the `select_related` here ensures that the N+1 issue will not happen if the calling code changes.
And there is no much benefit in prefetching it anyways, cause we score a single batch only.
So how about we get rid of the prefetch in the task instead?

PR #263, opened 2024-10-01
Set trusted miner & s3 config in install_validator.sh script

emnoor-reef:


README.md
```diff
@@ -58,7 +58,24 @@ for miner in miners:
...
+export TRUSTED_MINER_KEY=...
+export TRUSTED_MINER_ADDRESS=...
+export TRUSTED_MINER_PORT=...
+
+export S3_BUCKET_NAME_PROMPTS=...
+export S3_BUCKET_NAME_ANSWERS=...
+
+export AWS_ACCESS_KEY_ID=...
+export AWS_SECRET_ACCESS_KEY=...
+export AWS_ENDPOINT_URL=...
```
mpnowacki-reef, line 75: It's cool to have this configurable but is the sane default, i.e. what do I put here if I want to use plain s3?

emnoor-reef: Made `AWS_ENDPOINT_URL` optional. boto3 automatically uses the default endpoint for the given bucket.

PR #275, opened 2024-10-10
Move receipt models to library project

kkowalski-reef:
This PR migrates Receipt models to a different app -> old receipts will be gone. Apparently this is OK.

### DIFF
--- a/compute_horde/compute_horde/receipts/__init__.py
+++ b/compute_horde/compute_horde/receipts/__init__.py
@@ -1,1 +1,12 @@
+from compute_horde.receipts.schemas import Receipt, ReceiptType
+from compute_horde.receipts.transfer import ReceiptFetchError, get_miner_receipts
+
 default_app_config = "compute_horde.receipts.apps.ComputeHordeReceiptsConfig"
+
+# Reexported for compatibility. These were moved to submodules.
+__all__ = [
+    "Receipt",
+    "ReceiptType",
+    "get_miner_receipts",
+    "ReceiptFetchError",
+]


miner/app/src/compute_horde_miner/miner/miner_consumer/validator_interface.py
```diff
@@ -7,7 +7,10 @@
 
 import bittensor
 from compute_horde.mv_protocol import miner_requests, validator_requests
-from compute_horde.mv_protocol.validator_requests import BaseValidatorRequest
+from compute_horde.mv_protocol.models import JobFinishedReceipt, JobStartedReceipt
+from compute_horde.mv_protocol.validator_requests import (
+    BaseValidatorRequest,
+)
```
emnoor-reef, line 13: Does ruff not like simple formatting? :|
```suggestion
from compute_horde.mv_protocol.validator_requests import BaseValidatorRequest
```

compute_horde/compute_horde/settings.py
emnoor-reef, line 1: This module should be excluded from packaging. `compute_horde.settings` being importable is weird. IDK how it will interact with mypy if excluded though.

emnoor-reef (changes requested):
`compute_horde.mv_protocol` being a django app does not feel right to me. The protocol messages should not be part of a django app.

It could be `compute_horde.receipts` or something. We could move `receipts.py` module into this package. Then in near future we could implement optimized receipt scraping in this package.

All of these will also be used in facilitator. All the more reason not to name it `mv_`.

PR #282, opened 2024-10-15
added is_organic field to job started receipts

kkowalski-reef:


compute_horde/compute_horde/mv_protocol/validator_requests.py
```diff
@@ -133,6 +133,7 @@ class JobStartedReceiptPayload(ReceiptPayload):
     executor_class: ExecutorClass
     time_accepted: datetime.datetime
     max_timeout: int  # seconds
+    is_organic: bool = False
```
emnoor-reef, line 136: I don't think we should set a default here. We are already doing big breaking changes to the receipts, so it should not matter compatibility-wise.

emnoor-reef (approved):

PR #283, opened 2024-10-16
add huggingface volume type

andreea-popescu-reef:


executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -316,6 +318,19 @@ def __init__(self, concurrency=3, max_retries=3):
         self.semaphore = asyncio.Semaphore(concurrency)
         self.max_retries = max_retries
 
+    async def download_from_huggingface(
```
mzukowski-reef, line 321: sync to async?

executor/app/src/compute_horde_executor/settings.py
```diff
@@ -43,6 +43,8 @@ def wrapped(*args, **kwargs):
 
 ENV = env("ENV", default="prod")
 
+HF_ACCESS_TOKEN = env("HF_ACCESS_TOKEN", default=None)
```
mzukowski-reef, line 46: we should pass this in executor manager - as a reference implementation

mzukowski-reef (changes requested):

mzukowski-reef (approved):

PR #292, opened 2024-10-29
Split celery tasks

kkowalski-reef:
- Replaced single-container celery setup with one container per queue, one master celery process per container and multiple worker processes per master process. This makes it so that there's one top-level process in the container, and we avoid `celery multi` which, by spawning celery with `--detach`, makes it sometimes hang on startup: https://github.com/celery/celery/discussions/8141
- Added new env var: `CELERY_CONCURRENCY`, replacing `CELERY_MASTER_CONCURRENCY` and `CELERY_WORKER_CONCURRENCY`
- Moved `manage.py migrate` from celery's entrypoint script to the main django app`s entrypoint so that multiple celery processes don't try to migrate the DB at the same time.


emnoor-reef (changes requested):

validator/envs/prod/docker-compose.yml
```diff
@@ -67,8 +67,84 @@ services:
...
+    image: compute_horde_validator/app
+    init: true
+    healthcheck:
+      test: celery -A compute_horde_validator status > /dev/null || exit 1
+    restart: unless-stopped
+    env_file: ./.env
+    environment:
+      - DEBUG=off
+      - PROMETHEUS_MULTIPROC_DIR=/prometheus-multiproc-dir/celery-worker-weights
+    command: ./celery-entrypoint.sh -Q weights -E -l INFO -c ${CELERY_CONCURRENCY}
```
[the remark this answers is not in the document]
emnoor-reef: You also need to provide `-n` flag with unique names (may be same as queue names?) for them to properly "mingle".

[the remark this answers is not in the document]
emnoor-reef: Though I am not sure. Our current celery config does not set names and the workers do not mingle. I don't know what it means though. 🤷‍♂️

validator/app/envs/prod/celery-entrypoint.sh
```diff
@@ -1,33 +1,16 @@
 #!/bin/sh
 set -eu
 
-type prometheus-cleanup.sh && ./prometheus-cleanup.sh
+CELERY_CONCURRENCY=${CELERY_CONCURRENCY:-2}
+CELERY_LOGLEVEL=${CELERY_LOGLEVEL:-INFO}
+CELERY_MAX_TASKS_PER_CHILD=${CELERY_MAX_TASKS_PER_CHILD:-10}
```
olzhasar-reef, line 6: Is that an informed decision?
The default value of 10 feels quite small, restarting the worker after every 10 tasks might be an unnecessary overhead.
Do we even have memory problems somewhere?

validator/envs/runner/data/docker-compose.yml
```diff
@@ -48,38 +48,57 @@ services:
...
     tmpfs: /run
     depends_on:
       - redis
-      - db
     logging:
       <<: *logging
-    labels:
-      - "com.centurylinklabs.watchtower.enable=true"
+
+  celery-worker-weights:
```
olzhasar-reef, line 68: Are we really sure we want that many dedicated workers?
Commit and reveal operations happen just once per tempo, the worker will just sit silently the rest of the time occupying a good memory portion.
And there is no downscaling either
It's probably reasonable to have separate workers for synthetic jobs and llm generation, but I'm not so sure about everything else. Going from 2 workers to 5 all of a sudden is quite a big jump

olzhasar-reef (approved):

validator/app/envs/prod/celery-entrypoint.sh
olzhasar-reef: Leaving this for you to decide, I feel like something around 30 would more reasonable

emnoor-reef (approved):

PR #293, opened 2024-10-31
Improve LLM system events

emnoor-reef:
I am a bit concerned about querying the `SystemEvent` model. It can be a really big table if validators don't prune it regularly.

### DIFF
--- a/validator/app/src/compute_horde_validator/settings.py
+++ b/validator/app/src/compute_horde_validator/settings.py
@@ -402,9 +402,19 @@ if env.bool("HTTPS_REDIRECT", default=False) and not DEBUG:
 else:
     SECURE_SSL_REDIRECT = False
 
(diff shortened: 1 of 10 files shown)


emnoor-reef:
In a single synthetic job run, it will be called once in the `_generate_jobs` stage.

Even if it is called only once, I am concerned about the query latency, if `SystemEvent` table contains millions of rows. It can delay dispatching synthetic jobs dramatically between validators.

With the querys in prompt generation/answering tasks that happen every 5 minutes, we don't want to take much of that time.

Maybe I will sit on this task a little bit more before committing to something.

mzukowski-reef:
lets wait and discuss it - I am really concerned - we do some really "stupid" things because the chart was broken... I fixed this chart :D :P

mzukowski-reef (approved):

PR #298, opened 2024-11-04
Added another initial job request barrier

adalc-reef:
Should mitigate any slowness because of receipt signing or model serialization (hashcat inline volume) @emnoor-reef @olzhasar-reef

### DIFF
--- a/validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
+++ b/validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
@@ -258,6 +258,7 @@ class Job:
     exception_stage: str | None = None
 
     accept_barrier_time: datetime | None = None
+    accept_barrier_2_time: datetime | None = None
     accept_before_sent_time: datetime | None = None
     accept_after_sent_time: datetime | None = None
     accept_response: V0AcceptJobRequest | V0DeclineJobRequest | None = None
@@ -371,6 +372,7 @@ class Job:
             exception_time=_datetime_dump(self.exception_time),
             exception_stage=self.exception_stage,
             accept_barrier_time=_datetime_dump(self.accept_barrier_time),
+            accept_barrier_2_time=_datetime_dump(self.accept_barrier_2_time),
             accept_before_sent_time=_datetime_dump(self.accept_before_sent_time),
             accept_after_sent_time=_datetime_dump(self.accept_after_sent_time),
             accept_response=_model_dump(self.accept_response),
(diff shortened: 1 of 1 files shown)


validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -900,32 +902,47 @@ async def _generate_jobs(ctx: BatchContext) -> None:
...
-        timeout_seconds=job.job_generator.timeout_seconds(),
-        volume=job.volume if job.job_generator.volume_in_initial_req() else None,
-        job_started_receipt_payload=job.job_started_receipt_payload,
-        job_started_receipt_signature=job.job_started_receipt_signature,
-    )
-    request_json = request.model_dump_json()
+        if job is not None:
+            job.accept_barrier_2_time = datetime.now(tz=UTC)
 
     async with asyncio.timeout(max_spin_up_time):
```
olzhasar-reef, line 947: If we used `finally` above without `return` in the `except` block, the current part will be reached no matter what, and probably crash unnecessarily.

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -900,32 +902,47 @@ async def _generate_jobs(ctx: BatchContext) -> None:
...
-    job = ctx.jobs[job_uuid]
-    job.accept_barrier_time = barrier_time
-    client = ctx.clients[job.miner_hotkey]
-
-    _generate_job_started_receipt(ctx, job)
-    assert job.job_started_receipt_payload is not None
-    assert job.job_started_receipt_signature is not None
+    job: Job | None = None
+    try:
+        await start_barrier.wait()
```
olzhasar-reef, line 913: I would add hard timeouts around these barriers, just in case. Normally it shouldn't be needed, but feels much robust if there is a recovery mechanism. Something conceptually like the following comes to mind:

```py3
try:
    async with asyncio.timeout(HARD_LIMIT):
        await barrier_1.wait()
except asyncio.TimeoutError:
    log_error()
    barrier_1.abort()
    return
except asyncio.BrokenBarrierError:
    return  # to avoid spamming with exceptions

try:
    do_stuff()
except Exception as exc:
    log_error()
    return  # this will be run after awaiting barrier_2
finally:
    # handle the same way as with the first barrier
    try:
        async with asyncio.timeout(HARD_LIMIT):
            await barrier_2.wait()
    except asyncio.TimeoutError:
        log_error()
        barrier_2.abort()
    except asyncio.BrokenBarrierError:
        return

do_more_stuff()
```

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
olzhasar-reef: Oh, you're right. For some reason I thought everything below is the continuation of the finally block.

PR #300, opened 2024-11-05
Metrics

mpnowacki-reef:


validator/app/src/compute_horde_validator/validator/metrics.py
```diff
@@ -23,9 +28,25 @@ def metrics_view(request):
     if os.environ.get(ENV_VAR_NAME):
         registry = prometheus_client.CollectorRegistry()
         RecursiveMultiProcessCollector(registry)
+        registry.register(CustomCeleryCollector())
```
emnoor-reef, line 31: why register twice?

emnoor-reef (approved):

validator/app/src/compute_horde_validator/validator/metrics.py
mpnowacki-reef: I've copied if from cookiecutter template but I see now it is redundant

PR #303, opened 2024-11-06
more llm sys events

andreea-popescu-reef:
- added more detailed system events for llm jobs
- do not mark workload finished if a prompt was not answered


validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1214,6 +1214,15 @@ def llm_prompt_generation():
         return
 
     logger.info("There are %s series in the db, generating prompts", num_prompt_series)
+    SystemEvent.objects.create(
+        type=SystemEvent.EventType.LLM_PROMPT_GENERATION,
+        subtype=SystemEvent.EventSubType.PROMPT_GENERATION_STARTED,
+        long_description="",
+        data={
+            "prompt_series_count": unprocessed_workloads_count,
```
agoncharov-reef, line 1222: I suppose that `unprocessed_workloads_count` will always be zero

agoncharov-reef (approved):

PR #323, opened 2024-11-15
Add certificate generation functions

emnoor-reef:


mpnowacki-reef:
Two notes:
1. I think you don't need a CA in our case, you can generate a key and cert and not sign it any way
2. I don't think you need to set CN to anything, AFAIK certificate pinning makes the client disregard CN etc

mpnowacki-reef:
about testing. the endgoal is to start an nginx using this cert, pass the public part to some client (like python reqeusts) configured to trust that and only that cert and make sure that:
1. https connection is established (it can be hello world)
2. connection epecting a different cert (client side) fails (client side)

so this should be merged once a test like that is written (part of the CI pipeline or at least a manual one) to prove it work. The stuff about no need for CA or CN might be wrong, but a test would answer these questions.

mpnowacki-reef (changes requested):
posted as general comments

emnoor-reef:
I removed CA. But I couldn't make it work without CN or SAN, so kept them there, but used simpler values.

Added tests that spawns a nginx docker container in the tests. :')

mpnowacki-reef (approved):

PR #325, opened 2024-11-20
Better validator startup

mzukowski-reef:


mzukowski-reef:
works well on startup and even on stopping validator if is stopped by executing `docker-compose down` inside runner - if stopped by stoping runner with `--remove-orphans`, containers are stopped in random order and errors about redis connections are still possible, but we should not hide them

andreea-popescu-reef (approved):

PR #326, opened 2024-11-21
A6000 README instructions

mateuszsrebrny-reef:
In this PR:
- [x] describing `AWS_DEFAULT_REGION` variable
- [x] mention about A6000 being needed, and that the buckets needs to handle small text files

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ for miner in miners:
 
 ## Validator
 
-To run a validator, first you need to [setup a trusted miner for cross-validation](/validator/README.md#setting-up-a-trusted-miner-for-cross-validation) and [provision S3 buckets for prompts and answers](/validator/README.md#provision-s3-buckets-for-prompts-and-answers) of LLM jobs.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -77,10 +77,13 @@ export S3_BUCKET_NAME_ANSWERS=...
...
 export AWS_ACCESS_KEY_ID=...
 export AWS_SECRET_ACCESS_KEY=...
-export AWS_ENDPOINT_URL=...
+export AWS_DEFAULT_REGION=...
 ```
 
-Note: setting `AWS_ENDPOINT_URL` is optional. If not given, the default AWS S3 endpoint will be used.
+Note: `AWS_DEFAULT_REGION` property is optional. Use it when your buckets are not in your default AWS region.
+
+Export `AWS_ENDPOINT_URL` too if you want to us another S3 provider. If not given, the default AWS S3 endpoint will be used.
```
mpnowacki-reef, line 85:
```suggestion
Export `AWS_ENDPOINT_URL` too if you want to us another cloud object storage (s3-compatible) provider. If not given, AWS S3 will be used.
```

PR #328, opened 2024-11-25
Streamflow

andreea-popescu-reef:


compute_horde/compute_horde/certificate.py
```diff
@@ -9,6 +14,50 @@
 from cryptography.x509 import Certificate
 from cryptography.x509.oid import NameOID
 
+logger = logging.getLogger(__name__)
+
+
+def start_nginx_with_certificates(
```
mpnowacki-reef, line 20: where will this be called? Based on that it should either be async or not, hard to tell rn

compute_horde/compute_horde/certificate.py
```diff
@@ -9,6 +14,50 @@
...
+            "-p",
+            f"{port}:443",
+            "-v",
+            f"{tmp_path}:/etc/nginx/",
+            "nginx:1.26-alpine",
+        ]
+    )
+
+    # wait for nginx to start
+    time.sleep(1)
```
mpnowacki-reef, line 57: maybe it's better to do a request to see if server is up?

compute_horde/compute_horde/mv_protocol/miner_requests.py
```diff
@@ -50,6 +51,14 @@ class V0ExecutorReadyRequest(BaseMinerRequest, JobMixin):
     message_type: RequestType = RequestType.V0ExecutorReadyRequest
 
 
+# TODO: sign this with the miners wallet when sent to vali
+class V1ExecutorReadyRequest(V0ExecutorReadyRequest):
```
mpnowacki-reef, line 55: miners implementing their own executor managers will have to somehow provide data for this model, like the ip and port, right? Have you got an idea of how that interface will look like?

compute_horde/compute_horde/certificate.py
```diff
@@ -9,6 +14,50 @@
...
+
+    generate_certificate_at(certs_dir, "127.0.0.1")
+
+    nginx_conf_file = tmp_path / "nginx.conf"
+    nginx_conf_file.write_text(nginx_conf)
+
+    client_cert_file = certs_dir / "client.crt"
+    client_cert_file.write_bytes(public_key)
+
+    subprocess.run(
```
mpnowacki-reef, line 40: since this nginx will have to communicate with the job, maybe some networking shenanigans is required here? I'm not sure.

compute_horde/compute_horde/certificate.py
andreea-popescu-reef: yes - should be async

compute_horde/compute_horde/mv_protocol/miner_requests.py
andreea-popescu-reef: yes - will repost it together with the other pr

compute_horde/compute_horde/certificate.py
andreea-popescu-reef: we'll need to change the job to use a bridge network (or something more restricted - not sure how it would work if we keep network none?) then it should be able to connect from nginx/executor host

miner/app/src/compute_horde_miner/miner/executor_manager/executor_port_dispenser.py
```diff
@@ -0,0 +1,56 @@
...
+        self.available_ports = set(range(start_port, end_port))
+
+    def _is_port_available(self, port: int) -> bool:
+        """
+        Checks if a specific port is available on the system.
+        """
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            try:
+                s.bind(("", port))
```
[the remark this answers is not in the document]
andreea-popescu-reef: 👎

executor/app/src/compute_horde_executor/settings.py
```diff
@@ -266,6 +266,8 @@ def wrapped(*args, **kwargs):
 }
 
 MINER_ADDRESS = env.str("MINER_ADDRESS")
+NGINX_PORT = env("NGINX_PORT", default=6006)
```
mpnowacki-reef, line 269: this is a crucial part of the executor manager interface. Miners who implement their won need to know that this value needs to be passed to executors via env vars as they are started

miner/app/src/compute_horde_miner/miner/executor_manager/executor_port_dispenser.py
```diff
@@ -0,0 +1,56 @@
+import socket
+
+
+class ExecutorPortDispenser:
```
mpnowacki-reef, line 4: it should be noted that this isn't for production use. It's in `executor_manager` module, ergo part of the miner and therefore finds a free port on the miner instance and that only works if executors run on the same instance which is usually not the case in production. In production, executors get this port via an env var from the executor manager, before they are started.

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -64,6 +72,46 @@
...
+        listen 443 ssl;
+        server_name localhost;
+
+        ssl_certificate /etc/nginx/ssl/certificate.pem;
+        ssl_certificate_key /etc/nginx/ssl/private_key.pem;
+
+        ssl_client_certificate /etc/nginx/ssl/client.crt;
+        ssl_verify_client on;
+
+        location /execute-job {
```
mpnowacki-reef, line 88: well actually according to the streaming spec there is no spec for the routes in jobs, so `/*` needs to be routed to `/*` and it's the partie's making a request reposnsibility to know the route (After all they are the one providing the job image in the first place)

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -495,6 +559,26 @@ async def run_job(self, job_request: V0JobRequest):
...
+        if self.executor_certificate is not None:
+            # TODO wait for the streaming job to be ready
+            logger.debug("Spinning up nginx for streaming job")
+            await asyncio.sleep(1)
+            try:
+                # the job container ip is needed to proxy the request to the job container
+                job_ip = await get_docker_container_ip(self.job_container_name)
+                # update the nginx conf with the job container ip
+                nginx_conf = NGINX_CONF.replace("CONTAINER", f"{job_ip}:{JOB_CONTAINER_PORT}")
+                await start_nginx(
```
mpnowacki-reef, line 572: I think this function doesn't throw an error if getting a proper response to `/ok` from nginx fails and therefore the `except` will not trigger

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -495,6 +559,26 @@ async def run_job(self, job_request: V0JobRequest):
...
+                    nginx_conf,
+                    port=settings.NGINX_PORT,
+                    dir_path=self.nginx_dir_path,
+                    container_name=self.nginx_container_name,
+                )
+                logger.debug("Nginx started successfully")
+            except Exception as e:
+                logger.error(f"Failed to start Nginx: {e}")
```
mpnowacki-reef, line 580: shouldn't this trigger an error response back to the miner? the job won't work.

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -517,6 +601,20 @@ async def run_job(self, job_request: V0JobRequest):
             exit_status = process.returncode
             timeout = False
 
+        # if streaming job, stop the associated nginx server
```
mpnowacki-reef, line 604: just above there is

```
        except TimeoutError:
            # If the process did not finish in time, kill it
            logger.error(
                f"Process didn't finish in time, killing it, job_uuid={self.initial_job_request.job_uuid}"
            )
            process.kill()
            timeout = True
            exit_status = None
            stdout = (await process.stdout.read()).decode() if process.stdout else ""
            stderr = (await process.stderr.read()).decode() if process.stderr else ""
        else:
             exit_status = process.returncode
             timeout = False
             
```

which might be right for the first streaming job we're doing, which is supposed to kill itself after solving one request, but for most others coming to an end of life is natural and no cause for alarm. I guess It should be fixed, or at least a TODO and a ticket made

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -475,10 +539,10 @@ async def run_job(self, job_request: V0JobRequest):
             "run",
             *docker_run_options,
             "--name",
-            f"{settings.EXECUTOR_TOKEN}-job",
+            self.job_container_name,
             "--rm",
             "--network",
-            "none",
+            "bridge",  # to allow the container to access the nginx server
```
mpnowacki-reef, line 545: and this won't allow people knowing the ip address to access this thing directly, right?

miner/app/src/compute_horde_miner/miner/executor_manager/executor_port_dispenser.py
andreea-popescu-reef: so the dispenser would be used when the executors are on the same machine as the miner. If they rewrie the miner/executor-manager then this is indeed not useful.
In production / with executors on different instances the env var is not mandatory / it should be fine for all executors to fallback on the default port
is this just about documentation? or are you suggesting a code change?

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
andreea-popescu-reef: ... there is a issue with the health checks not working if executor is on a different machine than miner. health nginx is set to accept only connections from the executor machine now - will fix to permit connections from miner ip

miner/app/src/compute_horde_miner/miner/executor_manager/executor_port_dispenser.py
mpnowacki-reef: just documentation, a doctoring here and a note about the envvar in the docs, for people who implement their own executor managers (which ideally would be everybody)

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
mpnowacki-reef: the most important use case is the issuer of the job - they will primarily hit that nginx. nginx has to accept requests from all ips

mpnowacki-reef: except for the healthcheck. executor needs to do the health check and it won't have the ssl certs. maybe it can hit the job container directly, I'm not sure

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
andreea-popescu-reef: https://reeftechnologies.atlassian.net/browse/COM-376

compute_horde/compute_horde/mv_protocol/miner_requests.py
```diff
@@ -50,6 +51,14 @@ class V0ExecutorReadyRequest(BaseMinerRequest, JobMixin):
     message_type: RequestType = RequestType.V0ExecutorReadyRequest
 
 
+# TODO: sign this with the miners wallet when sent to vali
```
mpnowacki-reef, line 54: is there a jira ticket for it so we don't forget?

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -64,6 +72,51 @@
...
+            proxy_pass http://$upstream;
+        }
+    }
+
+    server {
+        listen 80;
+        server_name localhost;
+
+        # Allow access only from the miner instance
+        allow {MINER_ADDRESS};
```
mpnowacki-reef, line 101: is there a use case for the miner to reach the job?

mpnowacki-reef (changes requested):
comments left

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
andreea-popescu-reef: iirc there was a heartbeat logic for streaming jobs for miner

compute_horde/compute_horde/certificate.py
```diff
@@ -53,22 +53,27 @@ async def start_nginx(
...
     await process.wait()
 
     # wait for nginx to start
     ip = await get_docker_container_ip(container_name)
+    url = f"http://{ip}/ok"
     nginx_started = False
     for _ in range(WAIT_FOR_NGINX_TIMEOUT):
         try:
-            if requests.get(f"http://{ip}/ok").status_code == 200:
+            if requests.get(url).status_code == 200:
```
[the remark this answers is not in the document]
andreea-popescu-reef: fixed

executor/app/src/compute_horde_executor/settings.py
andreea-popescu-reef: added

miner/app/src/compute_horde_miner/miner/miner_consumer/executor_interface.py
```diff
@@ -69,18 +69,39 @@ async def connect(self):
...
+        else:
+            raise ValueError(f"Unknown job message type {request_type}")
+
+        await self.send(miner_initial_job_request.model_dump_json())
+
+    def get_executor_ip(self):
+        for key, value in self.scope["headers"]:
+            if key.decode("utf-8").lower() == "x-real-ip":
+                return value.decode("utf-8")
+        # Fallback to client's IP if header is not present
```
mpnowacki-reef, line 103: why would this address be ever accurate? it's gonna be the validator's addres, right?

miner/app/src/compute_horde_miner/miner/miner_consumer/executor_interface.py
```diff
@@ -91,6 +112,17 @@ async def handle(self, msg: BaseExecutorRequest):
             self.job.status = AcceptedJob.Status.FAILED
             await self.job.asave()
             await self.send_executor_failed_to_prepare(self.executor_token)
+        if isinstance(msg, executor_requests.V0StreamingJobReadyRequest):
+            await self.send_streaming_job_ready(
```
mpnowacki-reef, line 116: I'm not exactly sure, but it does look like the job status should be updated here, looking at other cases

executor/app/src/compute_horde_executor/executor/tests/integration/test_main_loop.py
```diff
@@ -111,6 +112,59 @@ def test_main_loop():
...
+                        "docker_run_cmd": ["python", "./mock_streaming_job.py", "autostart"],
+                        "docker_run_options_preset": "none",
+                        "job_uuid": job_uuid,
+                    }
+                ),
+            ]
+        )
+    )
+    command.handle()
+    assert [json.loads(msg) for msg in command.miner_client_for_tests.transport.sent_messages] == [
```
mpnowacki-reef, line 143: well if it's an integration test then i think a request should be to the job

mpnowacki-reef (changes requested):

miner/app/src/compute_horde_miner/miner/miner_consumer/executor_interface.py
mpnowacki-reef: ah no, this is executor interface, so this should be correct

miner/app/src/compute_horde_miner/miner/miner_consumer/executor_interface.py
```diff
@@ -69,18 +69,39 @@ async def connect(self):
...
+        elif request_type == validator_requests.RequestType.V1InitialJobRequest.value:
+            request = validator_requests.V1InitialJobRequest(**job.initial_job_details)
+            miner_initial_job_request = miner_requests.V1InitialJobRequest(
+                job_uuid=request.job_uuid,
+                base_docker_image_name=request.base_docker_image_name,
+                timeout_seconds=request.timeout_seconds,
+                volume=request.volume,
+                volume_type=request.volume_type.value if request.volume_type else None,
+                public_key=request.public_key,
+                executor_ip=self.get_executor_ip(),
```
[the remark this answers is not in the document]
andreea-popescu-reef: that's right that I think can be refactored out - however the miner can just send the apropiate executor ip (from the headers) back to the validator, there's no need for the executor to pass them in the msg

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -762,9 +912,29 @@ async def _executor_loop(self):
...
+                        job_ready = await check_endpoint(
+                            f"http://{ip}:{JOB_CONTAINER_PORT}/health",
+                            WAIT_FOR_STREAMING_JOB_TIMEOUT,
+                        )
+                        if job_ready:
+                            await miner_client.send_streaming_job_ready(
+                                certificate=job_runner.executor_certificate
+                            )
+                        else:
+                            await miner_client.send_streaming_job_failed_to_prepare()
```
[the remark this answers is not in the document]
andreea-popescu-reef: no, it should get cleaned

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -762,9 +912,29 @@ async def _executor_loop(self):
                     await miner_client.send_failed_to_prepare()
                     return
                 logger.debug(f"Running job {initial_message.job_uuid}")
-                result = await job_runner.run_job(job_request)
-                result.specs = specs
 
+                # start the job running process
+                result = await job_runner.start_job(job_request)
+                if result is None:
+                    # None result means this is a streaming job
```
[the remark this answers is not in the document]
andreea-popescu-reef: start_job will return a JobResult for regular jobs: https://github.com/backend-developers-ltd/ComputeHorde/pull/328/files/a362d25f9fe23d1d2080874122f2522256b2712c#diff-81e444d9966acde76c48aed02796c070e09e32a63aee793a266f9c30dd67e6a9R507 ?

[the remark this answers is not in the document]
andreea-popescu-reef: `# None result means this is a streaming job` that comment is wrong, the streaming job check happens after - fixed, refactored to be clearer

executor/app/src/compute_horde_executor/executor/tests/integration/test_main_loop.py
andreea-popescu-reef: that's testing more that the right messages are send back n forth - there is no request as the mock job is autostarted https://github.com/backend-developers-ltd/ComputeHorde/pull/328/files/a362d25f9fe23d1d2080874122f2522256b2712c#diff-75cf32fc45d7d468b4571fce4a11bce56cc526f8bf02523570e9f16dc9c1d693R29 
the actual "integration test" is the `debug_run_mock_streaming_job_to_miner` command

miner/app/src/compute_horde_miner/miner/miner_consumer/executor_interface.py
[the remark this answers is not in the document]
andreea-popescu-reef: reviewed it - executor ip is needed for certificate generation and fetching the public ip there and tagging it to the streaming messages seems cleaner (or I don't know a nice way for the executor to get it's own public ip)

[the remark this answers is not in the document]
mpnowacki-reef: I think the best thing to do in this case is to change the `ip` to `address` in this message and have the miner (specifically the executor manager) determine that value and send it to the executor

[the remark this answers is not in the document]
mpnowacki-reef: this way miners can explicitly declare it for each executor. And it can be different from the address the executor sees, it can be different from the address the miner reads from the socket and even from the address the miner uses to ssh to the executor to start the process. NATs and stuff. Load balancers and what not.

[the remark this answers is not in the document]
andreea-popescu-reef: that would add extra logic for anyone who wants to write their own executor manager right?

mpnowacki-reef (approved):

miner/app/src/compute_horde_miner/miner/miner_consumer/executor_interface.py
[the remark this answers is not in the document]
andreea-popescu-reef: fixed as discussed

miner/app/src/compute_horde_miner/miner/miner_consumer/executor_interface.py
andreea-popescu-reef: after the job starts the possible statuses are RUNNING, FAILED or FINISHED - if we add a STREAMING_READY status it would override the RUNNING status of the job
also both job prepare fail and whole job fail translate to FAILED job state - so streaming prep fail should result in a FAILED job state too - otherwise we can split the failures into FAILED, FAILED_PREPARE, FAILED_STREAMING but not sure if that helps much

mpnowacki-reef: you're right

mpnowacki-reef (approved):

PR #333, opened 2024-11-26
Remove TRUSTED_MINER_KEY

adal-chiriliuc-reef:


mpnowacki-reef (changes requested):
while this solves the issue I am afraid it will cause WTFisms in future readers of this code. And I think it can be avoided. The general idea is very good (use a fake key instead of the key optional for the miner client) I would suggest moving it to a different place. There are two use cases for this fake miner hotkey:

1. organicminerclient
2. LlmPromptsJobGenerator

for the first one I suggest making a subclass that doesn't expect miner key as an argument and pass the fake key to the super().__init__(...) call. Works the same but the logic is centralized in sth like `TrustedMinerClient`

for the seconds one the thing it's completely unused, `ain't` of LlmPromptJobGenerator is empty, it might as well not be called, or at least override to not expect this unnecessary argument.

mpnowacki-reef (approved):

PR #339, opened 2024-12-04
efficient receipts transfer

kkowalski-reef:
General ideas:
- Miners append serialized receipts to jsonl files, a file is equivalent to a "page" of receipts.
- Page files are append-only.
- Receipts are always appended to the "current" page.
- Current page number is determined by global clock so that we don't need to add additional communication about the pages themselves.
- Current page changes every 5 minutes, so as not to make the pages too big. (this may be tweaked or could be moved to a dynamic config)
- Validators keep up with receipts by continuously requesting 2 most recent pages - that's 2 to allow for slight desync between their clocks.
- Validators use Range request header to only retrieve receipts added since their last request - or in most cases get an empty response indicating there's nothing new to serve.
- When starting up, validators will request **all pages for the latest 5 hours** - so that they have 2 full cycles worth of data to work with.


compute_horde/compute_horde/receipts/transfer.py
```diff
@@ -1,108 +1,215 @@
...
                 continue
-
-            if receipt.payload.miner_hotkey != hotkey:
-                logger.warning(f"Miner sent receipt of a different miner {receipt=}")
-                continue
-
-            if not receipt.verify_miner_signature():
-                logger.warning(f"Invalid miner signature of receipt {receipt=}")
+            except BadMinerReceiptSignature as e:
+                logger.warning(
```
mpnowacki-reef, line 202: I'm wondering if `warning` is the right level. Maybe it should be info? It's a problem in external data, not necessary "our" problem

compute_horde/compute_horde/receipts/transfer_checkpoints.py
```diff
@@ -0,0 +1,42 @@
...
+                "Required settings.py setting missing: RECEIPT_TRANSFER_CHECKPOINT_CACHE"
+            )
+
+        self.cache = caches[settings.RECEIPT_TRANSFER_CHECKPOINT_CACHE]  # type: ignore
+
+    async def get(self, key: str) -> int:
+        try:
+            return int(await self.cache.aget(key, 0))
+        except TypeError:
+            logger.warning(f"Django cache contained non-integer checkpoint value for {key}")
```
mpnowacki-reef, line 33: this should be an error propagated to sentry

miner/app/src/compute_horde_miner/miner/eviction.py
```diff
@@ -29,3 +30,14 @@ def evict_receipts() -> None:
     JobStartedReceipt.objects.filter(timestamp__lt=cutoff).delete()
     JobAcceptedReceipt.objects.filter(timestamp__lt=cutoff).delete()
     JobFinishedReceipt.objects.filter(timestamp__lt=cutoff).delete()
+
+    logger.info("Evicting old receipt pages")
+    store = LocalFilesystemPagedReceiptStore()
+    all_pages = store.get_available_pages()
+    cutoff_page = store.current_page_at(cutoff)
+    to_delete = [p for p in all_pages if p < cutoff_page]
+    for page in to_delete:
```
mpnowacki-reef, line 39: I can't help but think that this loop should be in the LocalFilesystemPagedReceiptStore class, am I wrong?

validator/app/src/compute_horde_validator/celery.py
```diff
@@ -31,12 +31,10 @@
     "compute_horde_validator.validator.tasks.reveal_scores": "weights",
     "compute_horde_validator.validator.tasks.set_scores": "weights",
     "compute_horde_validator.validator.tasks.do_set_weights": "weights",
-    # Receipts
-    "compute_horde_validator.validator.tasks.fetch_receipts": "receipts",
-    "compute_horde_validator.validator.tasks.fetch_receipts_from_miner": "receipts",
     # Misc
     "compute_horde_validator.validator.tasks.send_events_to_facilitator": DEFAULT_QUEUE,
     "compute_horde_validator.validator.tasks.fetch_dynamic_config": DEFAULT_QUEUE,
+    "compute_horde_validator.validator.tasks.archive_receipt_pages": DEFAULT_QUEUE,
```
mpnowacki-reef, line 37: is there a task like that in vali? I thought that was miner stuff

validator/envs/runner/data/docker-compose.yml
```diff
@@ -225,6 +218,21 @@ services:
     labels:
       - "com.centurylinklabs.watchtower.enable=true"
 
+  transfer-receipts:
```
mpnowacki-reef, line 221: I' wondering if we have a way of knowing that this process froze, maybe a prometheus metric inside of it that "ticks"?

validator/app/src/compute_horde_validator/validator/management/commands/transfer_receipts.py
```diff
@@ -0,0 +1,215 @@
...
+from django.utils import timezone
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = "Fetch receipts from miners"
+
+    def add_arguments(self, parser):
+        parser.add_argument("--miner-hotkey", type=str)
```
mpnowacki-reef, line 24: there should be a help text here saying that this option is for testing only or it should be named `--debug-miner-hotkey` or sth

validator/app/src/compute_horde_validator/validator/management/commands/transfer_receipts.py
```diff
@@ -0,0 +1,215 @@
...
+        semaphore: asyncio.Semaphore,
+    ):
+        """
+        Runs indefinitely and polls for changes in active pages every `interval`.
+        """
+        while True:
+            start_time = time.monotonic()
+            current_page = LocalFilesystemPagedReceiptStore.current_page()
+            pages = list(reversed(range(current_page - N_ACTIVE_PAGES + 1, current_page + 1)))
+            receipts, exceptions = await ReceiptsTransfer.transfer(
```
mpnowacki-reef, line 194: I'd add prometheus metrics here

PR #346, opened 2024-12-15
possible fix for undelivered channel layer messages via group_send

kkowalski-reef:
Sets the default django channels backend to newer `channels_redis.pubsub.RedisPubSubChannelLayer`

In case this causes issues, validator and miners can revert this via .env:
`CHANNELS_BACKEND=channels_redis.core.RedisChannelLayer`

(and restart afterwards)


andreea-popescu-reef (approved):
should add the `if trouble revert this via .env:
CHANNELS_BACKEND=channels_redis.core.RedisChannelLayer` line somewhere - but I guess on discord not docs

PR #347, opened 2024-12-16
validator enable streaming synthetic jobs

andreea-popescu-reef:


validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -1382,32 +1453,82 @@ async def _multi_send_initial_job_request(ctx: BatchContext) -> None:
...
+            if isinstance(job.job_generator, LlmPromptsJobGenerator):
+                seed = job.job_generator.seed
+            else:
+                logger.error(f"Bad streaming job generator type: {job.job_generator}")
+                return
+
+            # wait to trigger all streaming jobs at the same time
+            await streaming_start_barrier.wait()
+
+            url = f"https://{response.ip}:{response.port}/execute-job"
```
mpnowacki-reef, line 1481: did you manage to make the job image succesfully autokill itself AFTER the response to the first `execute-job` request is delivered to the client? I tried doing that but it flaky, sometimes Flask died too early and that's why i introduced a second endpoint for terminating the process. Btw the response contains a hash of the file that will be sent to S3 and we have to remember (probably via a jira ticket) to later on store it here somewhere and compare it with what is read from S3

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -1382,32 +1453,82 @@ async def _multi_send_initial_job_request(ctx: BatchContext) -> None:
...
+                    if r.status_code == 200:
+                        logger.debug(
+                            f"Successfully triggered job execution for {job_uuid} on {url}"
+                        )
+                    else:
+                        raise Exception(
+                            f"Bad response triggering streaming job {job_uuid} execution {url}: {r.status_code}, {r.text}"
+                        )
+                except Exception as e:
+                    raise Exception(
```
mpnowacki-reef, line 1496: does this get propagated to sentry? IT should only be a system event

mpnowacki-reef (changes requested):
I see no explicit tests of the streaming flow. Are there any?

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
andreea-popescu-reef: yeah, modified the solver to stop at execute-job with a delay and testing if that's too flaky - can just mass send terminate requests though

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
andreea-popescu-reef: it will get caught up the chain and tracked as part of the send_job system events

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
andreea-popescu-reef: will add request to terminate

andreea-popescu-reef:
> I see no explicit tests of the streaming flow. Are there any?

https://github.com/backend-developers-ltd/ComputeHorde/pull/328#discussion_r1904331670  the `debug_run_mock_streaming_job_to_miner` command is testing the streaming flow and should be refactored in some form of intergration test (iirc we had some integration test framework that we ended up disabled? - so not sure how to refactor this command)

the streamin flow as part of synthetic jobs is a bit more involved to mock

mpnowacki-reef (approved):

PR #351, opened 2024-12-17
feature flag for receipts transfer

kkowalski-reef:
Safety killswitch just in case.
- Adds `DYNAMIC_RECEIPT_TRANSFER_ENABLED` "feature flag"
- When disabled: disables validators' transfer attempts, disables miners storing receipts to files
- The validators' transfer "daemon" will still run, but do nothing until the flag is enabled again

Additionally:
- Moved debug receipt generation task to miner as it would require adding contance to compute_horde
- "Current receipt store" will be decided by the app (miner/validator) instead of the library.
- Transfer procedure strongly assumes the pages receipt store implementation for now anyways

### DIFF
--- a/miner/app/src/compute_horde_miner/miner/dynamic_config.py
+++ b/miner/app/src/compute_horde_miner/miner/dynamic_config.py
@@ -1,13 +1,8 @@
 import asyncio
 import time
-from collections.abc import Callable
-from contextlib import suppress
-from typing import Any
 
 import constance.utils
 from asgiref.sync import sync_to_async
-from compute_horde.executor_class import ExecutorClass
-from constance import config
 from django.conf import settings
 
 
--- a/miner/app/src/compute_horde_miner/miner/eviction.py
+++ b/miner/app/src/compute_horde_miner/miner/eviction.py
@@ -2,9 +2,9 @@ import logging
 from datetime import timedelta
 
 from compute_horde.receipts.models import JobAcceptedReceipt, JobFinishedReceipt, JobStartedReceipt
-from compute_horde.receipts.store.local import LocalFilesystemPagedReceiptStore
-from compute_horde_miner.miner.receipts import current_store
 from django.utils.timezone import now
+
+from compute_horde_miner.miner.receipts import current_store
 
 from .models import AcceptedJob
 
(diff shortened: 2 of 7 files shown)


miner/app/src/compute_horde_miner/miner/dynamic_config.py
```diff
@@ -0,0 +1,35 @@
+import asyncio
+import time
+from collections.abc import Callable
+from contextlib import suppress
+from typing import Any
+
+import constance.utils
+from asgiref.sync import sync_to_async
+from compute_horde.executor_class import ExecutorClass
```
andreea-popescu-reef, line 9: unused?

andreea-popescu-reef: isn't the DynamicConfigHolder duplicated now? can't it stay in the computehorde lib?

kkowalski-reef: I didn't want to stuff in another requirement to the library - this requires constance.
I'll do another PR later to pull this up to compute_horde.

miner/app/src/compute_horde_miner/miner/eviction.py
```diff
@@ -31,6 +32,7 @@ def evict_receipts() -> None:
     JobAcceptedReceipt.objects.filter(timestamp__lt=cutoff).delete()
     JobFinishedReceipt.objects.filter(timestamp__lt=cutoff).delete()
 
-    logger.info("Evicting old receipt pages")
-    store = LocalFilesystemPagedReceiptStore()
-    store.delete_pages_older_than(cutoff)
+    store = current_store()
+    if isinstance(store, LocalFilesystemPagedReceiptStore):
```
andreea-popescu-reef, line 36: but why not skip the if and have all noopstore methods do nothing?

or alternatively make current_store return None and simplify the check here? then we would get rid of NoopStore and BaseStore too

andreea-popescu-reef (approved):

miner/app/src/compute_horde_miner/miner/eviction.py
kkowalski-reef: Fair - I'll add an `evict` method to the base store and do nothing in noopstore.evict
I don't want to return None because this way I will have to check for None everywhere I want to push a receipt to the store, and the code around there is already complex.

miner/app/src/compute_horde_miner/miner/dynamic_config.py
mpnowacki-reef: I do agree that adding constance as a dependancy to the compute_horde lib might be unfortunate but OTOH so is code duplication, so maybe injection is the answer? The `DynamicConfigHolder` could be instantiated like

```
DynamicConfigHolder(settings.CONSTANCE_CONFIG, constance.utils.get_values)
```

mpnowacki-reef:
There are no tests validator side regarding the feature flag, are there?

PR #353, opened 2024-12-20
Add block number cache

slawomir-gorawski-reef:
Issue: https://reeftechnologies.atlassian.net/browse/COM-353.

I created a new Django app for it, `compute_horde.blockchain`, so that other apps can easily use the background task to update the cache by just adding it to their installed apps. Can be moved somewhere else if needed.

Tested with unit tests and manually by adding this app and background task to the validator's settings.

Measured speed – block number from cache takes ~0.002 s, from subtensor ~0.1 s.

Usage:

* Add `compute_horde.blockchain` to INSTALLED_APPS in the settings module.
* Add `compute_horde.blockchain.tasks.update_block_cache` as a celery beat task with some reasonable frequency (like 1 or 2 s)
* To get the current block number from cache, use `aget_current_block` (or `get_current_block`).

### DIFF
--- a/compute_horde/compute_horde/blockchain/block_cache.py
+++ b/compute_horde/compute_horde/blockchain/block_cache.py
@@ -1,4 +1,8 @@
+import asyncio
+import functools
+
 import bittensor
+from asgiref.sync import async_to_sync
 from django.conf import settings
 from django.core.cache import cache
 
@@ -8,22 +12,40 @@ _BLOCK_CACHE_KEY = getattr(
 _BLOCK_CACHE_TIMEOUT = getattr(settings, "COMPUTE_HORDE_BLOCK_CACHE_TIMEOUT", 2)
 
 
-def get_subtensor(network):
+class BlockNotInCacheError(KeyError):
+    pass
+
+
+@functools.cache
+def _get_subtensor(network):
     return bittensor.subtensor(network)
 
 
-def get_current_block() -> int:
+async def aget_current_block(timeout: float = 1.0) -> int:
+    """
+    Gets the current block number from cache. Waits for ``timeout`` seconds if it's not there.
+
+    :param timeout: Number of seconds to wait for the block. Pass 0 to return or raise an exception immediately instead.
+    :raise BlockNotInCacheError: If block was not found in cache.
+    """
+    # Note: cache.get here is not async.
     current_block: int = cache.get(_BLOCK_CACHE_KEY)
     if current_block is not None:
         return current_block
 
-    return cache_current_block()
+    if timeout <= 0:
+        raise BlockNotInCacheError(_BLOCK_CACHE_KEY)
+
+    await asyncio.sleep(timeout)
+
+    return await aget_current_block(timeout=0)
 
 
-def cache_current_block() -> int:
-    subtensor = get_subtensor(network=settings.BITTENSOR_NETWORK)
-    current_block: int = subtensor.get_current_block()
+get_current_block = async_to_sync(aget_current_block)
+
+
+def cache_current_block() -> None:
+    subtensor = _get_subtensor(network=settings.BITTENSOR_NETWORK)
+    current_block = subtensor.get_current_block()
 
     cache.set(_BLOCK_CACHE_KEY, current_block, _BLOCK_CACHE_TIMEOUT)
-
-    return current_block
--- a/compute_horde/compute_horde/blockchain/block_cache.py
+++ b/compute_horde/compute_horde/blockchain/block_cache.py
@@ -1,0 +1,29 @@
+import bittensor
+from django.conf import settings
+from django.core.cache import cache
+
+_BLOCK_CACHE_KEY = getattr(
+    settings, "COMPUTE_HORDE_BLOCK_CACHE_KEY", "compute_horde.blockchain.block_cache.current_block"
+)
+_BLOCK_CACHE_TIMEOUT = getattr(settings, "COMPUTE_HORDE_BLOCK_CACHE_TIMEOUT", 2)
+
+
+def get_subtensor(network):
+    return bittensor.subtensor(network)
+
+
+def get_current_block() -> int:
+    current_block: int = cache.get(_BLOCK_CACHE_KEY)
+    if current_block is not None:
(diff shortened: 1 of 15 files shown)


compute_horde/compute_horde/blockchain/block_cache.py
```diff
@@ -0,0 +1,29 @@
...
+def get_current_block() -> int:
+    current_block: int = cache.get(_BLOCK_CACHE_KEY)
+    if current_block is not None:
+        return current_block
+
+    return cache_current_block()
+
+
+def cache_current_block() -> int:
+    subtensor = get_subtensor(network=settings.BITTENSOR_NETWORK)
```
mpnowacki-reef, line 24: this takes ~ 0.5s, is there any reason to get a new instance of the subtensor everytime?

compute_horde/pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
     "requests>=2.32.2",
     "Django~=4.2.4",
     "cryptography>=42.0.8",
+    "celery>=5.3.0",
```
mpnowacki-reef, line 17: there should be a limit, sth like `<6` cause even `from celery import shared_task` might break with a major release, right?

compute_horde/compute_horde/settings.py
```diff
@@ -1 +1,6 @@
-INSTALLED_APPS = ["compute_horde.receipts"]
+INSTALLED_APPS = [
```
mpnowacki-reef, line 1: what is this file for anyway?

compute_horde/uv.lock
```diff
@@ -281,6 +302,25 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/01/27/2b6460f1fa1772d23acaac3d4d5b4e311fb5fd7fc034064ddc1b4b374c5d/bt_decode-0.2.0a0-cp311-none-win_amd64.whl", hash = "sha256:bc0229921b993948fc1aacc2c83cae8c343f39d71ab8befa77d94b12ab7183b7", size = 322955 },
 ]
 
+[[package]]
+name = "celery"
+version = "5.5.0rc3"
```
mpnowacki-reef, line 307: is this correct? we want a release candidate to be used in production?

mpnowacki-reef (changes requested):
I really like your code, it's nice and concise. One thing lacking here is: a description on how this app is supposed to be used (and it's use in current projects: miner and vali). should projects call `get_current_block()`? is this the one and only public interface method? What about the task? should it be added to  beat config?

One design decicion i don't agree with is `return cache_current_block()` in `get_current_block()` (provided this is the public method that should be used by projects. This is IO that can be idefinitely blocking. Also it's not suitable to be used in async contexts. BTW is there a nice way to read the cached key (which i guess is in redis) in async functions? If the cache is empty then this function should wait fo a set period of time for it to be populated, maybe schedule a task for ti to happen (but do the subtensor IO blockingly) and raise an exception if it doesn't get populated. Don't block the whole app.

compute_horde/uv.lock
slawomir-gorawski-reef: Good catch, no. We have `prerelease = "allow"` in this project's uv settings, I think for bt-decode, but it causes uv to resolve dependencies to unstable releases. I managed to fix it manually but it can happen again if we keep it as is.

compute_horde/compute_horde/settings.py
slawomir-gorawski-reef: Tooling – at least django-stubs needs it, it seems.

compute_horde/compute_horde/blockchain/block_cache.py
slawomir-gorawski-reef: No reason, fixed.

slawomir-gorawski-reef:
Yeah, good point about async, I didn't think of that. I changed the main function to be async with a configurable timeout that raises an exception if the timeout is exceeded.

Reading from the cache itself is not async and Django doesn't support that yet.

It's Redis and it seems fast, so maybe it's fine to leave it as is?
Otherwise, I think the best solution would be to ditch `django.core.cache` and use an async Redis client from redis-py directly.

compute_horde/uv.lock
mpnowacki-reef: Please create a ticket in jira

mpnowacki-reef:
yeah i guess sync redis access is fine

mpnowacki-reef:
These points

>Usage:
>
>    Add compute_horde.blockchain to INSTALLED_APPS in the settings module.
>    Add compute_horde.blockchain.tasks.update_block_cache as a celery beat task with some reasonable frequency (like 1 or 2 s)
>    To get the current block number from cache, use aget_current_block (or get_current_block).

should be somewhere in the readme of this module or some docstring or some place in the repo.

compute_horde/compute_horde/blockchain/block_cache.py
```diff
@@ -0,0 +1,51 @@
...
+    settings, "COMPUTE_HORDE_BLOCK_CACHE_KEY", "compute_horde.blockchain.block_cache.current_block"
+)
+_BLOCK_CACHE_TIMEOUT = getattr(settings, "COMPUTE_HORDE_BLOCK_CACHE_TIMEOUT", 2)
+
+
+class BlockNotInCacheError(KeyError):
+    pass
+
+
+@functools.cache
```
mpnowacki-reef, line 19: this is gonna be primarily used by a celery task. Does `functools.cache` work nicely with celery tasks?

slawomir-gorawski-reef: Yup, I checked it, works as expected

slawomir-gorawski-reef:
Added to README.

> BTW can beat schedule tasks every second?

It's supported. As for performance, on my (weak) VM the tasks are completed within ~0.1s and the Redis queue seems very fast.
It doesn't have actual load from other tasks, but most of these are routed to separate queues and workers, so I'd say it should probably be still ok.

Screenshot 2025-01-02 at 11 19 52

mpnowacki-reef (approved):

PR #355, opened 2025-01-03
Add task to fetch validators in compute_horde

slawomir-gorawski-reef:
Issue: https://reeftechnologies.atlassian.net/browse/COM-392

As discussed with @mpnowacki-reef, we wanted to have the logic of updating the validators database implemented in compute_horde, to be used in stats collector, miner, facilitator and validator.

Migrating all of these projects to a new Validator model sounded a bit tough, so I made this a configurable Django app, where VALIDATOR_MODEL can be specified in settings, and it will use the model class of a given project, which all of them already have.

Code largely taken from miner, tested manually with the miner app models.

Not implemented yet: migrate miner to use this.


compute_horde/README.md
```diff
@@ -11,8 +14,42 @@ INSTALLED_APPS = [
...
+    ...,
+    'compute_horde.validators',
+    ...,
+]
+```
+
+### Models
+
+This app works with models from other projects; it doesn't define its own.
+To use it, you need to specify `VALIDATOR_MODEL` in your `settings.py`,
```
mpnowacki-reef, line 39: I think for a setting used by an app and one that is defined in the project's settings.py `VALIDATOR_MODEL` is too short, has a chance of clashing with sth. i think all the settings entries used by an app should have a common prefix, like COMPUTE_HORDE_VALIDATOR_...

slawomir-gorawski-reef: Makes sense, added. I left the BITTENSOR_ ones as-is, because it seems that all our projects use them the same way and it's less work than repeating them. Do you think that makes sense?

mpnowacki-reef: yeah, sure

mpnowacki-reef (approved):

PR #356, opened 2025-01-03
Add s3 diagnostic management command

emnoor-reef:


validator/app/src/compute_horde_validator/validator/management/commands/check_s3_setup.py
```diff
@@ -0,0 +1,97 @@
...
+from requests import HTTPError
+
+from compute_horde_validator.validator import s3
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument("--aws-access-key-id")
```
mpnowacki-reef, line 19: these need clarification on how they are used (meaning you fall back to what is in your .env if you don't pass them)

mpnowacki-reef (changes requested):
This needs some mention in some README

mpnowacki-reef (approved):

PR #364, opened 2025-01-13
Various llm fixes

mpnowacki-reef:


validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -760,6 +763,16 @@ class _MinerClientFactoryProtocol(Protocol):
...
 
 
+class BatchConfig:
+    def __init__(self):
+        self.event_limits: LimitsDict | None = None
+        self.llm_answer_s3_download_timeout: float | None = None
+
+    async def populate(self):
+        self.event_limits = await get_system_event_limits()
+        self.llm_answer_s3_download_timeout = await aget_config('DYNAMIC_LLM_ANSWER_S3_DOWNLOAD_TIMEOUT_SECONDS')
```
kkowalski-reef, line 773: I think it's better to craete a factory function for this in the dynamic_config package:

```python
@dataclass
class BatchConfig:
    event_limits: LimitsDict
    llm_answer_s3_download_timeout: float

async def get_batch_config() -> 'BatchConfig':
    return BatchConfig(
        event_limits=await get_system_event_limits(),
        self.llm_answer_s3_download_timeout=await aget_config('DYNAMIC_LLM_ANSWER_S3_DOWNLOAD_TIMEOUT_SECONDS'),
    )
```

and then instead
```python
config = BatchConfig()
await config.populate()
```
you do 
```python
config = await get_batch_config()
```

mpnowacki-reef: i like your idea but i maybe should be a class method, that would be even better

PR #365, opened 2025-01-13
miner: decline jobs when busy

kkowalski-reef:


validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -439,10 +441,52 @@ def emit_telemetry_event(self) -> SystemEvent | None:
...
+    def get_valid_decline_excuse_receipts(self) -> list[Receipt]:
+        assert self.job_started_receipt_payload is not None
+        assert isinstance(self.accept_response, V0DeclineJobRequest)
+
+        # Use the time when the receipt for this job was generated as an excuse time cutoff time
+        job_request_time = self.job_started_receipt_payload.timestamp
+
+        # We need time leeway so that if the miner receives an organic job around the same time as the synthetic, a slight
+        # time difference caused by clock desync and network latencies doesn't cause the miner to lose score when they
+        # accept an organic job and get a synthetic job request a couple of seconds "from the past"
```
mpnowacki-reef, line 464: I thought we don't have this problem because the receipts used for excusing are sent with the "job declined message" and are also signed by validators

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -439,10 +441,52 @@ def emit_telemetry_event(self) -> SystemEvent | None:
...
+
+        valid_receipts: list[Receipt] = []
+        for receipt in self.accept_response.receipts:
+            if (
+                isinstance(receipt.payload, JobStartedReceiptPayload)
+                and receipt.payload.is_organic
+                and receipt.payload.miner_hotkey == self.miner_hotkey
+                and receipt.payload.job_uuid != self.uuid
+                and receipt.payload.validator_hotkey in self.ctx.allowed_validators_for_excuse
+                and receipt.validator_signature not in seen_receipts
```
mpnowacki-reef, line 478: If this is to deter presenting the same receipt twice then i think we should base it on job_uuid, signatures are not deterministic, signing the same payload twice yields different signatures

validator/app/src/compute_horde_validator/validator/synthetic_jobs/batch_run.py
```diff
@@ -1305,19 +1384,29 @@ def _emit_decline_or_failure_events(ctx: BatchContext) -> None:
...
         ):
-            logger.warning("%s refused", job.name)
-            job.system_event(
-                type=SystemEvent.EventType.MINER_SYNTHETIC_JOB_FAILURE,
-                subtype=SystemEvent.EventSubType.JOB_NOT_STARTED,
-                description="refused",
-            )
+            if job.excused:
+                logger.warning("%s excused", job.name)
+                job.system_event(
```
mpnowacki-reef, line 1389: I think we should add info on which valis' receipts were provided to achieve this excuse so we can keep track of it and detect collusion

miner/app/src/compute_horde_miner/miner/executor_manager/v1/__init__.py
```diff
@@ -25,7 +25,7 @@
     "MAX_EXECUTOR_TIMEOUT",
     "BaseExecutorManager",
     "ExecutorClassPool",
-    "NoExecutorUnavailable",
+    "ExecutorUnavailable",
```
mpnowacki-reef, line 28: Wait isn't the name of this exception a part of the public interface?

miner/app/src/compute_horde_miner/miner/miner_consumer/validator_interface.py
```diff
@@ -348,16 +356,51 @@ async def handle_initial_job_request(
             executor_address = await current.executor_manager.get_executor_public_address(executor)
             job.executor_address = executor_address
             await job.asave()
-
+            await self.send(
+                miner_requests.V0AcceptJobRequest(job_uuid=msg.job_uuid).model_dump_json()
+            )
         except ExecutorUnavailable:
+            await self.group_discard(token)
+            await job.adelete()
```
mpnowacki-reef, line 364: we're deleting it? not saving how we excused?

mpnowacki-reef (changes requested):
I think there is not test checking that miner actually responds with `V0DeclineJobRequest`

miner/app/src/compute_horde_miner/miner/executor_manager/v1/__init__.py
kkowalski-reef: Yep. 
I reverted the usage of ExecutorUnavailable to what it was before my changes and instead introduced a new exception type for "All executors are busy".

miner/app/src/compute_horde_miner/miner/miner_consumer/validator_interface.py
kkowalski-reef: I mean, it's an "AcceptedJob" model. Does the miner need to remember excused jobs?

kkowalski-reef: Validator also has a job model and that one is marked as excused, is that what you mean?

PR #368, opened 2025-01-16
Temporary scoring formula

emnoor-reef:


validator/app/src/compute_horde_validator/settings.py
```diff
@@ -274,6 +274,21 @@ def wrapped(*args, **kwargs):
...
+        float,
+    ),
+    "DYNAMIC_SCORE_ORGANIC_JOBS_LIMIT": (
+        -1,
+        "Maximum number of organic jobs each miner can get scores for. Negative value means unlimited.",
+        int,
+    ),
+    "DYNAMIC_REJECTED_SYNTHETIC_JOB_SCORE": (
+        1.0,
+        "Score for each properly rejected synthetic job",
```
mpnowacki-reef, line 289: ❤️ I was about to ask you to add that param, after talking with people

mpnowacki-reef (approved):
this doesn't include dancing because we've talked about it after this PR was submitted, right?

validator/app/src/compute_horde_validator/validator/scoring.py
```diff
@@ -65,41 +65,105 @@ def score_jobs(
...
+        for hotkey, score in executor_class_organic_scores.items():
+            executor_class_scores[hotkey] += score
+
+        # normalize scores for executor class weight
+        normalized_scores = normalize(executor_class_scores, executor_class_weight)
+        for hotkey, score in normalized_scores.items():
             batch_scores[hotkey] += score
+
+    # apply manifest bonus
+    previous_batch = SyntheticJobBatch.objects.order_by("-id").exclude(id=batch.id).first()
```
mpnowacki-reef, line 149: I can see it was like that before, meaning that "previous batch" was determined based on id order, but it somehow hurts me terribly. Do you know why it's based on id? We have block numbers, we can search for a batch from the actual previous cycle. Taking the largest id might yield unexpected results if a vali skips a cycle. What should happen then? i think it should be assumed there is no previous batch.

validator/app/src/compute_horde_validator/validator/tests/test_scoring.py
```diff
@@ -160,3 +170,172 @@ def test_score_batches_executor_classes_weights():
...
+    )
+    SyntheticJob.objects.create(
+        miner=miner2,
+        status="PROPERLY_REJECTED",  # TODO: fix after merging with Kordian's PR
+        score=1,
+        **common_params,
+    )
+
+    scores = score_batches([batch])
+    assert scores.get("hotkey1", 0) > 0
```
mpnowacki-reef, line 211: are these calculations tested elsewhere? Because this assertion will not detect unintended changes in the calculation algorithm, i think there should be a `== ...` here

validator/app/src/compute_horde_validator/validator/tests/test_utils.py
```diff
@@ -562,7 +416,7 @@ async def interaction_callback(miner_client, after_job_sent):
             lambda m, j=job: m.payload.job_uuid == str(j.job_uuid), V0JobFinishedReceiptRequest
         )[0]
         time_took = receipt.payload.time_took_us / 1_000_000
-        assert abs(job.score * time_took - expected_multiplier) < 0.0001
+        assert abs(job.score * time_took - 1) < 0.0001
```
mpnowacki-reef, line 419: after my changes that got merged it's possible that this assertion needs to change in order to not take `time_took` into account when calculating the score. Just use `1`

mpnowacki-reef (changes requested):

validator/app/src/compute_horde_validator/validator/scoring.py
emnoor-reef: > Do you know why it's based on id?

🤷‍♂️ 

> What should happen then? i think it should be assumed there is no previous batch.

Yeah, that sounds about right. When validators miss a batch, they should apply bonus on the next batch. I'll adjust it.

validator/app/src/compute_horde_validator/validator/tests/test_scoring.py
emnoor-reef: I've added `test_temporary_scoring_formula` which checks the formula for 1 executor class. IDK if the same test should also check executor class weights or not. There's already a test thats only checks for class weights, but not excused/organic jobs (`test_score_batches_executor_classes_weights`). I'm not sure if checking both together is necessary or not.

PR #371, opened 2025-01-17
convenience pre-commit script

kkowalski-reef:


precommit.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env sh
+
+set -e
+
+for projectdir in compute_horde executor miner validator; do
+  cd "${projectdir}"
+  uv sync --all-groups
```
mpnowacki-reef, line 7: this assumes all 4 projects use the same venv? it doesn't get timeconsuming to run `uv sync` in all of them? I wonder if sth can be done, for example .envrc to use a different venv per subproject without any hassle

kkowalski-reef: uv will notice the subproject context and execute within it. and uv sync is very fast.

PR #375, opened 2025-01-20
Clean up some sentry error spamming

adalc-reef:


mpnowacki-reef:
LGTM but i honestly don't know if it's accurate from the point of view of subtensor interface. Are these attributes of Exceptions part of the public API? can we count on them always being there?

mpnowacki-reef (approved):

mpnowacki-reef:
well if it works then it works

PR #378, opened 2025-01-23
Enable huggingface hf_transfer

adalc-reef:


executor/app/envs/prod/Dockerfile
```diff
@@ -41,6 +41,7 @@ RUN apt-get update \
 WORKDIR /root/src/
 ENV PYTHONUNBUFFERED=1
 ENV PATH="/root/src/.venv/bin:$PATH"
+ENV HF_HUB_ENABLE_HF_TRANSFER=1
```
emnoor-reef, line 44: do you need it here, since you are already setting `os.environ` in python?

emnoor-reef (approved):

miner/pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
     "prometheus-client~=0.17.0",
     "django-prometheus==2.3.1",
     "django-business-metrics @ git+https://github.com/reef-technologies/django-business-metrics.git@9d08ddb3a9d26e8a7e478110d7c8c34c3aa03a01",
-    "django-constance[database]==3.1.0",
+    "django-constance==3.1.0",
```
mpnowacki-reef, line 27: why was this removed?

PR #382, opened 2025-01-24
Prepare executor for SN21 integration

slawomir-gorawski-reef:
Changes summary:

- Increase executor timeouts (that's required as 6 min is too low for SN21 jobs, it needs at least 12–15 min on A6000). I'm open to adjusting the values or doing it some better way.
- Add Adal's Huggingface download optimization env var to dev executor class
- Add more logging to executor (that includes job's stdout and stderr, if that's too much I can remove it)
- Add more logging to executor manager, including when executors are killed, it wasn't clear to me what's happening and why

### DIFF
--- a/compute_horde/compute_horde/executor_class.py
+++ b/compute_horde/compute_horde/executor_class.py
@@ -68,8 +68,6 @@ EXECUTOR_CLASS = {
 }
 
 
-# we split 144min 2 tempos window to 24 validators - this is total time after reservation,
-# validator may wait spin_up time of executor class to synchronize running synthetic batch
 MAX_EXECUTOR_TIMEOUT = timedelta(minutes=20).total_seconds()
 
 DEFAULT_EXECUTOR_CLASS = ExecutorClass.spin_up_4min__gpu_24gb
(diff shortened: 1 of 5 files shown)


compute_horde/compute_horde/executor_class.py
```diff
@@ -70,7 +70,7 @@ class ExecutorClassSpec:
 
 # we split 144min 2 tempos window to 24 validators - this is total time after reservation,
 # validator may wait spin_up time of executor class to synchronize running synthetic batch
-MAX_EXECUTOR_TIMEOUT = timedelta(minutes=6).total_seconds()
+MAX_EXECUTOR_TIMEOUT = timedelta(minutes=20).total_seconds()
```
emnoor-reef, line 73: I think the above comment became very confusing after this change. IMO removing it would be better.

emnoor-reef (approved):

compute_horde/compute_horde/executor_class.py
slawomir-gorawski-reef: Yeah, I agree, I was gonna do it but I forgot. Thanks!

PR #384, opened 2025-01-27
Allow specifying files to download in HuggingfaceVolume

slawomir-gorawski-reef:
For SN21 integration we want to be able download some files from a huge dataset repo without having to download the entire repository. I added two things to make it possible:

*  `repo_type`: must be `"dataset"` for dataset repositories, otherwise it doesn't work. Can be left as `None` for our current usage (models).
* `allow_patterns`: this is passed as-is to the hugginface_hub function. I use it to specify the list of files that I need, but I left the original signature (`str | list[str]`) because we might want to specify a single pattern at some point, I saw something like that in the SDK design issue.

### DIFF
--- a/compute_horde/compute_horde/base/volume.py
+++ b/compute_horde/compute_horde/base/volume.py
@@ -23,9 +23,12 @@ class VolumeType(str, enum.Enum):
 class HuggingfaceVolume(pydantic.BaseModel):
     volume_type: Literal[VolumeType.huggingface_volume] = VolumeType.huggingface_volume
     repo_id: str
+    # Set to "dataset" or "space" for a dataset or space, None or "model" for a model.
     repo_type: str | None = None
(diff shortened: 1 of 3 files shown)


compute_horde/compute_horde/base/volume.py
```diff
@@ -23,8 +23,10 @@ def __str__(self):
 class HuggingfaceVolume(pydantic.BaseModel):
     volume_type: Literal[VolumeType.huggingface_volume] = VolumeType.huggingface_volume
     repo_id: str
+    repo_type: str | None = None
     revision: str | None = None  # Git revision id: branch name / tag / commit hash
     relative_path: str | None = None
+    allow_patterns: str | list[str] | None = None
```
mpnowacki-reef, line 29: an explanation would be very useful

mpnowacki-reef (approved):

PR #390, opened 2025-01-30
Fix stearming synthetic

mpnowacki-reef:


validator/app/src/compute_horde_validator/validator/tests/test_synthetic_jobs/test_multiple_miners.py
```diff
@@ -233,11 +267,22 @@ async def test_all_streaming_succeed(
...
-        await check_synthetic_job(job_uuid, miner.pk, SyntheticJob.Status.COMPLETED, MOCK_SCORE)
+        if job_uuid == job_uuids[-1]:
+            await check_synthetic_job(job_uuid, miner.pk, SyntheticJob.Status.FAILED, 0)
+        elif job_uuid == job_uuids[-2]:
+            await check_synthetic_job(
+                job_uuid,
+                miner.pk,
+                SyntheticJob.Status.FAILED,
+                0,
+                re.compile("took too long: time_took_sec=.*"),
```
slawomir-gorawski-reef, line 282: I understand this is because of the "sleepy request" in the prep method?
I'd move that setup to this test, if possible, IMO it'll be easier to understand.

mpnowacki-reef: you mean to place the body of the prep method inside of the test method? I'm thinking maybe the per miner configuration should be more explicit and be sth like


```
[
Miner(hotkey=..., ws_messages='all', http_responses='ok'),
Miner(hotkey=..., ws_messages='up_until_ready', http_responses='ok'),
Miner(hotkey=..., ws_messages='all', http_responses='timeout'),
...
]
```

which would make it even more readable and would avoid the reader to have bind associate a miner it's http mock config via port numbers and other nonsense. WDYT?

validator/app/src/compute_horde_validator/validator/synthetic_jobs/generator/llm_prompts.py
```diff
@@ -145,5 +166,90 @@ def verify(self, msg: V0JobFinishedRequest, time_took: float) -> tuple[bool, str
...
+                # send the seed to the executor to start the streaming job
+                try:
+                    t_before = time.time()
+                    r = await client.post(
+                        url, json={"seed": self.seed}, headers={"Host": server_address}
+                    )
+                except Exception as e:
+                    msg = f"Failed to execute streaming job {job_uuid} on {url}: {e}"
+                    logger.debug(msg)
+                else:
```
slawomir-gorawski-reef, line 222: I think early returns are easier to follow than `try/except/else`, if they're possible. I think we can do that in this case – we don't need to move further and request to `/terminate` the job that hasn't been successfully started, right?

Same for below

validator/app/src/compute_horde_validator/validator/synthetic_jobs/generator/llm_prompts.py
```diff
@@ -104,18 +116,27 @@ async def download_answers(self, client: httpx.AsyncClient | None = None):
...
         return True
 
+    async def streaming_preparation_timeout(self) -> float | None:
+        """For streaming jobs, the timeout between sending a JobRequest and receiving StreamingReadyRequest"""
+        # TODO: caching
+        val = await aget_config("DYNAMIC_SYNTHETIC_STREAMING_JOB_READY_TIMEOUT")
+        return float(val) if val is not None else None
+
 
 class LlmPromptsSyntheticJobGenerator(LlmPromptsJobGenerator):
```
slawomir-gorawski-reef, line 141: Note, maybe out of scope: the naming of these classes is not clear to me (`BaseSyntheticJobGenerator -> LlmPromptsJobGenerator -> LlmPromptsSyntheticJobGenerator`). Could the last one be called something with "streaming" instead, maybe?

validator/app/src/compute_horde_validator/validator/synthetic_jobs/generator/llm_prompts.py
```diff
@@ -37,6 +48,7 @@ def __init__(
         self.s3_output_bucket = settings.S3_BUCKET_NAME_ANSWERS
 
         self.prompt_answers: dict[str, str] = {}
+        self.streaming_processing_time: float | None = None
```
slawomir-gorawski-reef, line 51: I think this can be moved to the subclass, it'd be cleaner this way

validator/app/src/compute_horde_validator/validator/tests/test_synthetic_jobs/mock_generator.py
```diff
@@ -27,6 +27,12 @@
 
 
 class MockSyntheticJobGenerator(BaseSyntheticJobGenerator):
+    def verify_time(self, time_took: float) -> bool | None:
+        return True
+
+    def verify_correctness(self, msg: V0JobFinishedRequest) -> tuple[bool, str]:
+        return True, "mock"
```
slawomir-gorawski-reef, line 34: I think `verify` can be removed from this class now, and the `TimeTookScoreMockSyntheticJobGenerator` class and factory seem to no longer be not needed at all

slawomir-gorawski-reef (approved):

validator/app/src/compute_horde_validator/validator/tests/test_synthetic_jobs/test_multiple_miners.py
slawomir-gorawski-reef: Yes, I didn't necessarily mean all of it, but the "sleepy request" part.
It's that I had to go to the fixture to see how many uuids there are, then calculate the port number and then go to that function to see why job uuid[-2] is supposed to time out – having the setup in the test would make this easier, I think.

Your suggestion from this comment is even better, of course, if you think it's worth to spend time doing that.

validator/app/src/compute_horde_validator/validator/synthetic_jobs/generator/llm_prompts.py
mpnowacki-reef: i think we do need to terminate, or at least try

PR #400, opened 2025-02-07
Artifacts

adalc-reef:


executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -725,6 +740,18 @@ async def wait_for_job(self, job_request: V0JobRequest) -> JobResult:
...
 
+        artifacts = {}
+        for artifact_filename in os.listdir(self.artifacts_mount_dir):
+            artifact_path = self.artifacts_mount_dir / artifact_filename
+            if os.path.isfile(artifact_path):
+                with open(artifact_path, "rb") as f:
+                    content = f.read()
+                artifact_size = len(content)
+                if artifact_size < MAX_ARTIFACT_SIZE:
+                    artifacts[artifact_filename] = base64.b64encode(content).decode()
```
slawomir-gorawski-reef, line 751: I guess it's valid either way and wasn't specified, but I expected these to be absolute paths, that's what was suggested in the SDK design too ([COM-341](https://reeftechnologies.atlassian.net/browse/COM-341)). I think it'd be easier for me as a user to have the same thing in two places, in job `with open(ARTIFACT_PATH, 'wb') as f: f.write(...)` and then in client read from `artifacts[ARTIFACT_PATH]`, it being the same in both places.
What do you think?

slawomir-gorawski-reef (commented):
Looks good and works well, just some remarks that we should probably agree on before merging

slawomir-gorawski-reef (approved):
Looks good to me after the paths are changed

PR #413, opened 2025-02-12
Upgrade bittensor to v9

emnoor-reef:


emnoor-reef:
The executor failing test is unrelated to this change. It is fixed here: https://github.com/backend-developers-ltd/ComputeHorde/pull/414

PR #422, opened 2025-02-18
Route jobs to DEBUG_MINER if it's set in env

kkowalski-reef:


kkowalski-reef:
yolo-merging this as it's techically me pushing @mpnowacki-reef code after having reviewed it myself.

PR #423, opened 2025-02-18
Track huggingface download errors

adalc-reef:


compute_horde/compute_horde/em_protocol/executor_requests.py
```diff
@@ -15,6 +15,10 @@ class RequestType(enum.Enum):
     GenericError = "GenericError"
 
 
+class ErrorType(enum.StrEnum):
```
kkowalski-reef, line 18: Can we call this `ExecutorErrorType`? This gets imported all over the code and in amny places `ErrorType` without this context is ambiguous.

kkowalski-reef: (just the type name, `error_type` fields and variables are fine)

kkowalski-reef (changes requested):
Shouldn't this also be supported in synthetic jobs?

slawomir-gorawski-reef (approved):

PR #424, opened 2025-02-20
reuse subtensor instance to avoid a memory leak 

kkowalski-reef:
(and because it's probably a good idea)

### DIFF
--- a/validator/app/src/compute_horde_validator/validator/management/commands/transfer_receipts.py
+++ b/validator/app/src/compute_horde_validator/validator/management/commands/transfer_receipts.py
@@ -119,11 +119,10 @@ class Command(BaseCommand):
         else:
             # 3rd, if no specific miners were specified, get from metagraph.
             logger.info("Will fetch receipts from metagraph miners")
+            subtensor = bittensor.subtensor(network=settings.BITTENSOR_NETWORK)
 
             async def miners():
-                metagraph = bittensor.metagraph(
-                    netuid=settings.BITTENSOR_NETUID, network=settings.BITTENSOR_NETWORK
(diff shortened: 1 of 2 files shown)


validator/app/src/compute_horde_validator/validator/management/commands/transfer_receipts.py
```diff
@@ -119,11 +119,10 @@ async def miners():
         else:
             # 3rd, if no specific miners were specified, get from metagraph.
             logger.info("Will fetch receipts from metagraph miners")
+            subtensor = bittensor.subtensor(network=settings.BITTENSOR_NETWORK)
 
             async def miners():
-                metagraph = bittensor.metagraph(
-                    netuid=settings.BITTENSOR_NETUID, network=settings.BITTENSOR_NETWORK
-                )
```
emnoor-reef, line 126: So, the memory leak was creating lots of subtensor (that holds ws connections) and not properly closing them? That's a hard find! I think we need to go through the codebase and convert all the metagraph/subtensor calls to:

```python
with bittensor.Subtensor(...) as subtensor:
    metagraph = subtensor.metagraph(...)
```

Or hold on to the subtensor like here.

emnoor-reef (approved):

validator/app/src/compute_horde_validator/validator/management/commands/transfer_receipts.py
kkowalski-reef: Yes, and the amount of memory is significant with each call - it's a matter of 100s, maybe 1000s of calls and we're down

PR #427, opened 2025-02-21
Update is_active miner flag

adalc-reef:


miner/app/src/compute_horde_miner/miner/executor_manager/_internal/selector.py
```diff
@@ -114,18 +114,21 @@ async def active(self, hotkeys: Sequence[str], block: int = None) -> str | None:
...
-            cycle,
+        peak_cycle = get_peak_cycle(block=block, netuid=settings.BITTENSOR_NETUID)
+        peak_cycles = [
+            peak_cycle,
         ]
 
         for i in range(self._lookback):
-            cycles.append(
+            peak_cycles.append(
                 get_cycle_containing_block(
```
emnoor-reef, line 124: From the name, I assume all the cycles in `peak_cycles` list should be peak cycles.

```suggestion
                get_peak_cycle(
```

Though, to be honest, I don't really understand what is happening here with this list.

emnoor-reef (approved):

PR #436, opened 2025-02-25
add compute_horde lib depency to compute_horde_sdk

andreea-popescu-reef:


slawomir-gorawski-reef (changes requested):
As we discussed on Slack it may be better to avoid the compute-horde dependency in the SDK due to:

* strict bittensor version pinning – may be not good for other subnets that want to update sooner or later one than us
* transitive dependency on Django and Celery – may be a hard sell if other subnets don't use Django, and may cause resolution problems if they do

PR #437, opened 2025-02-25
Facilitator list[str] args

adalc-reef:


slawomir-gorawski-reef (approved):
If deleting the jobs in migration is ok then it looks good; I tested it with SN15 and it works

PR #459, opened 2025-03-05
refactor set_scores + bittensor 9.3.0 bump

andreea-popescu-reef:


kkowalski-reef (approved):

kkowalski-reef:
I think this needs more work?
- the `set_scores` celery task still calls subtensor
- the scores test is not mocked properly

kkowalski-reef (changes requested):
I think this needs more work -
- the `set_scores` celery task still calls subtensor
- the scores test is not mocked properly 
- possibly more?

PR #470, opened 2025-03-14
(WIP) move organic jobs to celery queue

kkowalski-reef:


validator/app/src/compute_horde_validator/validator/models.py
```diff
@@ -172,10 +173,18 @@ def get_latest(cls):
             raise Exception(msg)
         return metagraph
 
+    @classmethod
+    async def aget_latest(cls):
```
mpnowacki-reef, line 177: shouldn't this be typed? or does mypy figure out that it is an instance of `MetagraphSnapshot`?

validator/app/src/compute_horde_validator/validator/organic_jobs/facilitator_client.py
```diff
@@ -232,32 +234,47 @@ async def heartbeat(self):
                     )
             await asyncio.sleep(self.HEARTBEAT_PERIOD)
 
+    async def handle_job_status_updates(self):
+        class JobStatusChannelEnvelope(BaseModel):
```
mpnowacki-reef, line 238: why is this ca=lass defined in the method?

kkowalski-reef: I only need it here to deserialize / validate the dict. I don't want it to show up anywhere else.

kkowalski-reef: I'd say when refactoring the facilitator client at some point we will think about a good architecture for this.

PR #471, opened 2025-03-14
Check validator stake based on total effective stake

emnoor-reef:
Is this still correct: `MIN_STAKE = 1000`? @mateuszsrebrny-reef

### DIFF
--- a/compute_horde/compute_horde/utils.py
+++ b/compute_horde/compute_horde/utils.py
@@ -1,4 +1,5 @@
 import asyncio
+import dataclasses
 import datetime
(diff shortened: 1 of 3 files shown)


mateuszsrebrny-reef:
> Is this still correct: `MIN_STAKE = 1000`? @mateuszsrebrny-reef

yes, confirmed `1000 worth of ALPHAs` of weighted stake (meaning the one from `MetagraphInfo`, meaning sum of alphas and 0.18 of root taos)

mpnowacki-reef (approved):

PR #472, opened 2025-03-15
fix github action for releasing compute horde sdk

mpnowacki-reef:


compute_horde_sdk/pyproject.toml
```diff
@@ -29,17 +27,14 @@ dependencies = [
 "Issue Tracker" = "https://github.com/backend-developers-ltd/compute-horde-sdk/issues"
 
 [build-system]
-requires = ["pdm-backend==2.4.3"]
-build-backend = "pdm.backend"
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
```
emnoor-reef, line 31: I thought you needed a `setup.cfg` for `setuptools.build_meta ` to work. I guess they added support for reading metadata from `pyproject.toml`. But their docs are [still outdated](https://setuptools.pypa.io/en/latest/build_meta.html#how-to-use-it). :|

PR #477, opened 2025-03-19
Enable mypy strict mode for SDK

emnoor-reef:
- fix `nox -s type_check`
- add type checks in CI
- enable mypy strict mode
  - do related changes


compute_horde_sdk/src/compute_horde_sdk/_internal/models.py
```diff
@@ -59,18 +59,18 @@ class FacilitatorJobResponse(pydantic.BaseModel):
...
     # hf_repo_id: str
     # hf_revision: str
     # input_url: str
     # output_download_url: str
     # tag: str
     stdout: str
-    volumes: list = []
-    uploads: list = []
+    # volumes: list = []
+    # uploads: list = []
```
emnoor-reef, line 71: I've commented out volumes and uploads fields in FacilitatorJobResponse. The fields were not used anywhere, the class is not public, and I didn't want to try to figure out the generic params the lists 🤷‍♂️

PR #480, opened 2025-03-21
COM-484 Add miner signature to V0StreamingJobReadyRequest

emnoor-reef:


compute_horde/compute_horde/protocol_messages.py
```diff
@@ -132,6 +132,10 @@ class V0StreamingJobReadyRequest(BaseModel):
     public_key: str
     ip: str | None = None  # set by miner after it receives the message from executor
     port: int
+    miner_signature: str | None = None
+
+    def blob_for_signing(self) -> str:
+        return f"{self.ip}:{self.port}:{self.public_key}"
```
kkowalski-reef, line 138: Any reason not to include the job_uuid in the blob?

kkowalski-reef (approved):

compute_horde/compute_horde/protocol_messages.py
emnoor-reef: I missed it 😅

PR #486, opened 2025-03-24
COM-522 Ignore ResourceWarning for unclosed redis connections

emnoor-reef:
I suspected that there are some dangling asyncio tasks that are using redis/postgres connections. I fixed some, but the error still persists. Maybe my suspicions are unfounded and django keeps those connection open anyway? But it is curious that this issue only happens in validator, not in miner or facilitator.

As discussed, I've added ignore flags for these ports.

### DIFF
--- a/validator/app/src/compute_horde_validator/validator/tests/conftest.py
+++ b/validator/app/src/compute_horde_validator/validator/tests/conftest.py
@@ -134,3 +134,15 @@ def validators_with_this_hotkey(settings, validators):
 @pytest.fixture(scope="session")
 def run_uuid():
     return str(uuid.uuid4())
+
+
+# NOTE: Use this fixture when you need to find dangling asyncio tasks. It is currently commented
+#       because redis channels layers keeps dangling tasks, that makes the tests fail -_-
+# @pytest_asyncio.fixture(autouse=True)
+# async def check_still_running_tasks():
+#     yield
+#     tasks = asyncio.all_tasks()
+#     if len(tasks) > 1:
+#         raise ValueError(
+#             "\n" + "\n".join(f"{task.get_name()}: {task.get_coro()}" for task in tasks)
+#         )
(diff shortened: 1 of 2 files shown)


emnoor-reef:
Digging a bit deeper, it seems it only happens with redis, not postgres. Using this fixture:

```python
@pytest_asyncio.fixture(autouse=True)
async def check_still_running_tasks():
    yield
    tasks = asyncio.all_tasks()
    if len(tasks) > 1:
        raise ValueError(
            "\n" + "\n".join(f"{task.get_name()}: {task.get_coro()}" for task in tasks)
        )
```

We see this output like this for a bunch of tests:

```
ERROR compute_horde_validator/validator/tests/test_organic_jobs/test_miner_gets_blacklisted.py::test_miner_is_blacklisted__after_job_reported_cheated - ValueError: 
Task-176: <coroutine object RedisSingleShardConnection._do_receiving at 0x12095aea0>
Task-180: <coroutine object _wrap_asyncgen_fixture.<locals>._asyncgen_fixture_wrapper.<locals>.finalizer.<locals>.async_finalizer at 0x120c89a80>
```

The task is created here: https://github.com/django/channels_redis/blob/4.2.0/channels_redis/pubsub.py#L340

But I don't see [the `flush()` method](https://github.com/django/channels_redis/blob/4.2.0/channels_redis/pubsub.py#L281) being called from anywhere 🤷‍♂️

validator/app/src/compute_horde_validator/validator/tests/test_facilitator_client.py
```diff
@@ -266,7 +266,7 @@ async def test_facilitator_client__job_completed(ws_server_cls):
             facilitator_uri = f"ws://{host}:{port}/"
             facilitator_client = FacilitatorClient(get_keypair(), facilitator_uri)
 
-            async with ws_server.condition:
+            async with ws_server.condition, facilitator_client:
```
kkowalski-reef, line 269: Does this work? __aenter__ starts some tasks that could interfere with tests unless mocked

kkowalski-reef (approved):

validator/app/src/compute_horde_validator/validator/tests/test_facilitator_client.py
emnoor-reef: They seem to work. I'll revert these. I did not see your changes in the organic-in-celery PR before making these changes.

PR #491, opened 2025-04-03
Token-based authentication in Facilitator

piotr-figwer-reef:
This PR introduces a new authentication flow for the Facilitator. The existing request-by-request hotkey signature authentication is replaced with a `nonce -> sign -> login -> token` flow.

What changed:

1. SDK:
    * New `authenticate()` method has been added to the `ComputerHordeClient`.
    * SDK now caches the token in `self_token`.
    * Requests automatically triggers authentication if no valid token is present.
 
2. Facilitator:
   * `HotkeyAuthentication` is replaced with token-based `JWTAuthentication`.
   * `HotkeyWhitelist` is introduced to explicitly contril which hotkeys are permitted to obtain auth token. The list is configurable through the admin panel.


compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
```diff
@@ -213,6 +213,34 @@ def __init__(
...
+
+        login_url = urljoin(self.facilitator_url, "auth/login")
+        payload = {"hotkey": self.hotkey.ss58_address, "signature": signature, "nonce": nonce}
+
+        try:
+            login_response = await self._client.post(login_url, json=payload)
+        except httpx.HTTPError as e:
+            raise ComputeHordeError(f"Login request failed: {e}") from e
+
+        token = login_response.json().get("token")
```
mpnowacki-reef, line 239: both here and in case of the nonce I think it would be better to wrap `login_response.json().get("token")` in a try/except and throw a `ComputeHordeError` (or sth a little bit more specific to authentication?) in case the response is not a valid json or the top level json is not a dict.

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
```diff
@@ -232,7 +260,10 @@ async def _make_request(
         """
         :return: Response content as string.
         """
-        auth_headers = self._get_authentication_headers(method, urljoin(self.facilitator_url, url))
```
mpnowacki-reef, line 235: I think a lock would be a good thing here

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
```diff
@@ -232,7 +260,10 @@ async def _make_request(
...
+        if not self._token:
+            await self.authenticate()
+
+        auth_headers = self._get_authentication_headers()
```
mpnowacki-reef, line 266: no reauthorization?

facilitator/app/src/project/core/authentication.py
```diff
@@ -1,69 +1,55 @@
...
-        if not hotkey and not nonce and not signature:
-            return None
-
-        if not all((hotkey, nonce, signature)):
-            raise AuthenticationFailed("Missing some of required headers: hotkey, nonce, signature")
+def is_hotkey_allowed(hotkey: str) -> bool:
+    return HotkeyWhitelist.objects.filter(ss58_address=hotkey).exists() and \
+           (Validator.objects.filter(ss58_address=hotkey, is_active=True) or
+            Miner.objects.filter(ss58_address=hotkey, is_active=True)
+            )
```
mpnowacki-reef, line 18:
```suggestion
    return HotkeyWhitelist.objects.filter(ss58_address=hotkey).exists()
```

facilitator/pyproject.toml
```diff
@@ -46,6 +46,7 @@ dependencies = [
     "pyyaml>=6.0.1",
     "compute-horde",
     "compute-horde-sdk",
+    "PyJWT>=2.10.0"
```
mpnowacki-reef, line 49:
```suggestion
    "PyJWT>=2.10.0,<3.0.0"
```

facilitator/app/src/project/core/authentication.py
```diff
@@ -1,69 +1,55 @@
...
+
+        hotkey = payload.get("sub")
+        if not hotkey:
+            raise AuthenticationFailed("Token missing subject.")
+
+        if not is_hotkey_allowed(hotkey):
+            raise AuthenticationFailed("Unauthorized hotkey.")
 
         request.hotkey = hotkey
         return AnonymousUser(), hotkey
```
mpnowacki-reef, line 55: I'm not sure this is correct. I know it was here before, but it is a tiny bit of WTFism to successfully authenticate and then return `AnonymousUser()` as the user. WDYT?

mpnowacki-reef (changes requested):

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
piotr-figwer-reef: Yup, the lock is there now!

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
piotr-figwer-reef: Good point, I've added logic for handling expired tokens.

facilitator/app/src/project/core/authentication.py
piotr-figwer-reef: 👍

facilitator/pyproject.toml
piotr-figwer-reef: 👍

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
piotr-figwer-reef: Makes sense, it's wrapped in try/except block now.

facilitator/app/src/project/core/authentication.py
piotr-figwer-reef: Yeah, django user is not used in other parts of facilitator right now. But maybe it can be leveraged in the future? I've changed it, so now user is created/loaded after successful authentication for authenticated hotkey. Does it makes sense, or do you think we should create something custom?

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
```diff
@@ -213,6 +220,45 @@ def __init__(
...
+
+    async def authenticate(self) -> None:
+        nonce_url = urljoin(self.facilitator_url, "auth/nonce")
+        try:
+            response = await self._client.get(nonce_url)
+        except httpx.HTTPError as e:
+            raise ComputeHordeError(f"Nonce request failed: {e}") from e
+
+        try:
+            data = response.json()
```
mpnowacki-reef, line 234:
```suggestion
            data = response.json()
            nonce = data.get("nonce")
```

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
```diff
@@ -213,6 +220,45 @@ def __init__(
...
+        except Exception as e:
+            raise ComputeHordeError("Failed to parse nonce response") from e
+
+        nonce = data.get("nonce")
```
mpnowacki-reef, line 238:
```suggestion

```

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
```diff
@@ -213,6 +220,45 @@ def __init__(
...
+        login_url = urljoin(self.facilitator_url, "auth/login")
+        payload = {"hotkey": self.hotkey.ss58_address, "signature": signature, "nonce": nonce}
+
+        try:
+            login_response = await self._client.post(login_url, json=payload)
+        except httpx.HTTPError as e:
+            raise ComputeHordeError(f"Login request failed: {e}") from e
+
+        try:
+            data = login_response.json()
```
mpnowacki-reef, line 253:
```suggestion
            data = login_response.json()
            token = data.get("token")
```

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
```diff
@@ -213,6 +220,45 @@ def __init__(
...
+        except Exception as e:
+            raise ComputeHordeError("Failed to parse login response") from e
+
+        token = data.get("token")
```
mpnowacki-reef, line 257:
```suggestion

```

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
```diff
@@ -281,6 +300,11 @@ async def _make_request(
             response.raise_for_status()
         except httpx.HTTPStatusError as e:
             logger.error("Response status=%d: %s", e.response.status_code, e.response.text)
+            if e.response.status_code == 401 and "Token expired" in e.response.text:
+                async with self._token_lock:
+                    self._token = None
+                logger.info("Token expired, re-authenticating")
+                raise ExpiredTokenError("Token expired") from e
```
mpnowacki-reef, line 307: I see that this will be retried due to `@tenacity.retry(` but I can't help but think that it is slightly off, because if a request fails 3 times and then a token expires the user will get quite a backoff (exponential) just to authenticate and retry. I think the retry I really like the simplicity of the flow you've implemented, meaning no need for calling `authenticate()` - it's gonna happen on it's own. And that's worth keeping.but maybe the TokenExpired retry deserves a separate retry mechanism, with no backoff and not exhausting the `HTTP_RETRY_MAX_ATTEMPTS` count?

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
```diff
@@ -213,6 +220,45 @@ def __init__(
         self.facilitator_url = facilitator_url
         self._client = httpx.AsyncClient(base_url=self.facilitator_url, follow_redirects=True)
         self._signer = BittensorWalletSigner(hotkey)
+        self._token_lock = asyncio.Lock()
+        self._token: str | None = None
+
+    async def authenticate(self) -> None:
+        nonce_url = urljoin(self.facilitator_url, "auth/nonce")
+        try:
+            response = await self._client.get(nonce_url)
```
mpnowacki-reef, line 229: this doesn't throw on the server replying with a non-200 code, right? if that's true then I think both here and in `POST/login` there should be `raise_for_status`

facilitator/app/src/project/core/authentication.py
```diff
@@ -1,69 +1,59 @@
...
+            raise AuthenticationFailed("Token decode error.")
+
+        hotkey = payload.get("sub")
+        if not hotkey:
+            raise AuthenticationFailed("Token missing subject.")
+
+        if not is_hotkey_allowed(hotkey):
+            raise AuthenticationFailed("Unauthorized hotkey.")
+
+        user, created = User.objects.get_or_create(
```
mpnowacki-reef, line 51: yes this is great, I love this. The purist in me does, however, desire these users to be marked somehow. WDYT? maybe a boolean column? Or a relation to `HotkeyWhitelist`?

facilitator/app/src/project/core/authentication.py
mpnowacki-reef: Yes, In Django world, I think, being authenticated SHOULD also mean "related to some user" and creating one upon request is a very good strategy.

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
piotr-figwer-reef: 👍

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
piotr-figwer-reef: 👍

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
piotr-figwer-reef: 👍

compute_horde_sdk/src/compute_horde_sdk/_internal/sdk.py
piotr-figwer-reef: Makes sense, I've changed it so it has a separate retry mechanism now.

facilitator/app/src/project/core/authentication.py
piotr-figwer-reef: For simplicity, I'd keep this association on the `HotkeyWhitelist` side. This way, we will have the relationship marked without needing to customize the default Django `User` class.

piotr-figwer-reef:
Closing this PR in favor of [the one based on the main repository ](https://github.com/backend-developers-ltd/ComputeHorde/pull/494) branch so that all workflows can run.

PR #493, opened 2025-04-06
(WIP) Organic job puml

kkowalski-reef:


kkowalski-reef:
merging - we went through the diagram with @mpnowacki-reef

PR #501, opened 2025-04-18
Add retries to huggingface volume downloads, also with hf_transfer disabled

slawomir-gorawski-reef:
Our current SN15 jobs on ComputeHorde prod executors fail with errors like:
```
Failed to download model from Hugging Face: An error occurred while downloading using hf_transfer. Consider disabling HF_HUB_ENABLE_HF_TRANSFER for better error handling.
```

On the other hand, this library is useful when it works, because it can reduce the time to download a model by a lot.

Added retires to HF volume downloads with disabling `hf_transfer` after the first try.

It modifies an internal library constant, which is probably not a great solution, but I don't know how to do it better. Using or not `hf_transfer` is decided by an environment variable on import time and can't be parametrized otherwise.

<details><summary>Test script</summary>

```python
import os
from pathlib import Path

os.environ.setdefault("HF_HUB_ENABLE_HF_TRANSFER", "1")
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "compute_horde_executor.settings")

import django

django.setup()

import hf_transfer

def crash(*args, **kwargs):
    raise Exception("Crash test")

hf_transfer.download = crash

from compute_horde_executor.executor.management.commands.run_executor import DownloadManager

dm = DownloadManager()

dm.download_from_huggingface(Path("./testdir/"), "snx999/dev3", None)
```
</details>

### DIFF
--- a/executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
+++ b/executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
@@ -16,6 +16,7 @@ import typing
 import zipfile
 
 import httpx
+import huggingface_hub
 import packaging.version
 import pydantic
 from asgiref.sync import sync_to_async
@@ -57,7 +58,6 @@ from compute_horde_core.volume import (
 )
 from django.conf import settings
 from django.core.management.base import BaseCommand
-from huggingface_hub import snapshot_download
 from pydantic import TypeAdapter
 
 from compute_horde_executor.executor.output_uploader import OutputUploader, OutputUploadFailed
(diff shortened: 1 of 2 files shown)


executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -400,22 +400,34 @@ def download_from_huggingface(
...
+                    repo_id=repo_id,
+                    repo_type=repo_type,
+                    revision=revision,
+                    token=settings.HF_ACCESS_TOKEN,
+                    local_dir=relative_path,
+                    allow_patterns=allow_patterns,
+                )
+                return
+            except Exception as e:
+                logger.error(f"Failed to download model from Hugging Face: {e}")
```
mpnowacki-reef, line 418: hmmmm does this retry on non-existent repos as well?

mpnowacki-reef: Probably but then again that's a fast fail so that's okay probably

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -400,22 +400,34 @@ def download_from_huggingface(
...
+                last_exc = e
+                retries += 1
+                if huggingface_hub.constants.HF_HUB_ENABLE_HF_TRANSFER:
+                    # hf_transfer makes downloads faster, but sometimes fails where vanilla download doesn't.
+                    logger.info("Disabling hf_transfer for retry")
+                    huggingface_hub.constants.HF_HUB_ENABLE_HF_TRANSFER = False
```
mlech-reef, line 424: Is it safe to keep it `False` for every consecutive usage?

mlech-reef (approved):

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
mlech-reef: I don't mean retries, but e.g. the next download

mpnowacki-reef: the next download is gonna be in a separate executor process most likely

mpnowacki-reef: ah but it's different in your case @mlech-reef

mpnowacki-reef: the only way to circumvent this is to add a method for downloading from hf that has a lock (so only one would be running at a time) and has a context manager for setting this constant

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
slawomir-gorawski-reef: yeah, may be overkill but I thought the same

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
slawomir-gorawski-reef: is it important? I guess I can revert it back to original value after all retires

mpnowacki-reef: I think it's best @mlech-reef takes care of that in his PR? it will be easier for him to figure it out?

mlech-reef: I will, thanks

slawomir-gorawski-reef: Missed these messages, added my solution to restoring the orig value

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
slawomir-gorawski-reef: added a separate `except` for non-retryable errors

PR #504, opened 2025-04-28
Move upload/download logic from the executor to SDK (core)

mlech-reef:


mlech-reef:
Approved in slack
image

PR #506, opened 2025-04-30
Collateral smart contract integration

emnoor-reef:


validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1300,6 +1306,69 @@ def sync_metagraph() -> None:
             new_cycle_start_metagraph, snapshot_type=MetagraphSnapshot.SnapshotType.CYCLE_START
         )
 
+        hotkeys_to_sync = [n.hotkey for n in new_cycle_start_metagraph.neurons]
+
+        try:
+            sync_collaterals(subtensor, hotkeys_to_sync, current_cycle.start)
```
kkowalski-reef, line 1312: after merging with bt ddos shield, this will be a shielded metagraph. it's a subclass of the regular bittensor metagraph and should be compatible but i'd double check this on release.

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1300,6 +1306,69 @@ def sync_metagraph() -> None:
...
+    for miner in miners:
+        if not miner.uid:
+            continue
+
+        evm_address = associations.get(miner.uid)
+        if not evm_address:
+            continue
+
+        miner.evm_address = evm_address
+        to_update.append(miner)
```
kkowalski-reef, line 1348: is it possible that the address association disappears from the chain? because this will then keep the last known evm address stays in the database

validator/app/src/compute_horde_validator/validator/organic_jobs/facilitator_client.py
```diff
@@ -340,6 +341,22 @@ async def report_miner_cheated_job(self, job_uuid: str) -> None:
             },
         )
 
+        # slash collateral
```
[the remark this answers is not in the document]
kkowalski-reef: Technically a consumer can issue bogus jobs and keep slashing - nothing we can do about this, but having idempotency would guard against stupid mistakes.

validator/app/src/compute_horde_validator/validator/models.py
```diff
@@ -198,6 +200,9 @@ class Miner(models.Model):
     ip_version = models.IntegerField(default=4)
     port = models.IntegerField(default=0)
 
+    evm_address = models.CharField(max_length=42, default="")
+    collateral = models.FloatField(default=0)
```
kkowalski-reef, line 204: Again, storing currency as float feels wrong - why not use a Decimal field?

validator/app/src/compute_horde_validator/validator/models.py
```diff
@@ -198,6 +200,9 @@ class Miner(models.Model):
     ip_version = models.IntegerField(default=4)
     port = models.IntegerField(default=0)
 
+    evm_address = models.CharField(max_length=42, default="")
```
kkowalski-reef, line 203: Can we use a null as a default value? Makes it more explicit that an evm address may be absent.

kkowalski-reef (changes requested):
IMO the currency-as-float and multiple slashing prevention warrant some more work.

validator/app/src/compute_horde_validator/validator/tasks.py
emnoor-reef: AFAIK, they don't disappear. I also didn't see any way to remove an association once you've posted them to subtensor.

validator/app/src/compute_horde_validator/validator/organic_jobs/facilitator_client.py
```diff
@@ -340,6 +341,22 @@ async def report_miner_cheated_job(self, job_uuid: str) -> None:
...
+        if slash_amount > 0 and job.miner.evm_address:
+            try:
+                w3 = await collateral.get_async_web3_connection(network=settings.BITTENSOR_NETWORK)
+                async with w3:
+                    await collateral.slash_collateral(
+                        w3=w3,
+                        contract_address=settings.COLLATERAL_CONTRACT_ADDRESS,
+                        miner_address=job.miner.evm_address,
+                        amount_tao=slash_amount,
+                        url=f"job {job_uuid} cheated",
```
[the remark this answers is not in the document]
emnoor-reef: IDK. the smart contract calls it `url` 🤷‍♂️

validator/app/src/compute_horde_validator/validator/organic_jobs/facilitator_client.py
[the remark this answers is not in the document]
kkowalski-reef: how about a celery task? you could do a simple preliminary get from the db (not get for update) and check `if not slashed` in the ws message handler itself, then submit a task if it's not been slashed yet, then worst case some celery workers will be blocked for a couple of milliseconds.

kkowalski-reef:
@emnoor-reef sorry, I got one more thing :sweat_smile:  The whole "report cheated" function should only ever work for completed jobs - `OrganicJob.Status.COMPLETED`

validator/app/src/compute_horde_validator/validator/tasks.py
emnoor-reef: Moved check so that if evm_address doesn't exist on subtensor, it would set null to db.

kkowalski-reef (approved):

PR #507, opened 2025-04-30
receipts container needs the wallet now (bt shield)

kkowalski-reef:


kkowalski-reef:
closing; already merged elsewhere

PR #510, opened 2025-05-05
Fixing CI

kkowalski-reef:
uv released 0.7, nox broke in CI. This pins uv to an older (0.6) version as a workaround.


kkowalski-reef:
nvm, bumping nox instead.

PR #512, opened 2025-05-05
fix ci: bump nox

kkowalski-reef:


emnoor-reef (commented):
This probably also needs to be updated: https://github.com/backend-developers-ltd/ComputeHorde/blob/0d6528d19d4fffa4a2f6cb0a47d92b598963eeff/facilitator/pyproject.toml#L96

PR #513, opened 2025-05-05
Add logging in case of failed /tmp cleanup

slawomir-gorawski-reef:
We seem to be having problems with `/tmp` directories on executors not getting cleaned properly, added some logging to be able to see why

### DIFF
--- a/executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
+++ b/executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
@@ -716,11 +716,19 @@ class JobRunner:
             "sh",
             "-c",
             f"rm -rf {shlex.quote(root_for_remove.as_posix())}/*",
-            stdout=asyncio.subprocess.DEVNULL,
-            stderr=asyncio.subprocess.DEVNULL,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


mpnowacki-reef (approved):

executor/app/src/compute_horde_executor/executor/management/commands/run_executor.py
```diff
@@ -716,11 +716,19 @@ async def clean(self):
...
-            stdout=asyncio.subprocess.DEVNULL,
-            stderr=asyncio.subprocess.DEVNULL,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
         )
-        await process.wait()
-        self.temp_dir.rmdir()
+        stdout, stderr = await process.communicate()
+        if process.returncode == 0:
+            self.temp_dir.rmdir()
```
mpnowacki-reef, line 724: wait why not removing if the job fails?

slawomir-gorawski-reef: the "job" here is just `docker run rm -rf`, to be on the same page
if it fails then there're likely some files left over in that dir and `rmdir` would raise an exception
we can replace this with `shutil.rmtree` I guess, which at least would try, but it could also fail because of the file permissions

PR #514, opened 2025-05-06
Add miner dynamic config for job preload images

slawomir-gorawski-reef:


miner/pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
     "prometheus-client~=0.17.0",
     "django-prometheus==2.3.1",
     "django-business-metrics @ git+https://github.com/reef-technologies/django-business-metrics.git@9d08ddb3a9d26e8a7e478110d7c8c34c3aa03a01",
-    "django-constance==3.1.0",
+    "django-constance==4.1.3",
```
kkowalski-reef, line 27: Migration form 3->4 changes serialization of the config in the database from pickle to json. This makes the miner explode when loading the config page in django admin if there was already some config stored in the db. (was in my case locally)
 i don't think there's anything important in miner's constance config so purging the constance table should be fine.

kkowalski-reef: also, since we're already bumping this, let's go with latest, or is there are reason not to?

kkowalski-reef (changes requested):

kkowalski-reef (approved):
Right, should have checked for the migration first. My bad. We good.

PR #535, opened 2025-05-26
Route jobs based on allowance

emnoor-reef:


validator/app/src/compute_horde_validator/validator/organic_jobs/routing.py
```diff
@@ -80,19 +114,45 @@ async def pick_miner_for_job_v2(request: V2JobRequest) -> Miner:
     latest_miner_manifest: dict[str, MinerManifest] = {}
     for manifest in manifests:
         latest_miner_manifest[manifest.miner.hotkey] = manifest
-    manifests = list(latest_miner_manifest.values())
 
     # Discard manifests that explicitly say there are no executors of this type
-    manifests = [manifest for manifest in manifests if manifest.online_executor_count > 0]
```
andreea-popescu-reef, line 86: wouldn't it be more simple to add executor_class to the ComputeTimeAllowance table and query that to get the latest_miner_manifests along with the allowances directly?

emnoor-reef: IDK. Either way is fine. But as disscussed on the daily last week, I'll keep it out of this PR and work on that separately.

PR #545, opened 2025-06-02
feat: TurboBT

zyzniewski-reef:


compute_horde/compute_horde/utils.py
```diff
@@ -74,6 +79,46 @@ def get_validators(
...
+        validators.sort(
+            key=lambda validator: (
+                validator.hotkey == BAC_VALIDATOR_SS58_ADDRESS,
+                validator.stake,
+            ),
+            reverse=True,
+        )
+
+        try:
+            next(
```
andreea-popescu-reef, line 108: this seems like a convoluted way to handle the list containing BAC_VALIDATOR_SS58_ADDRESS
would be more readable with just a for loop?

PR #558, opened 2025-06-13
Add multi-GPU hardware support for LLM prompt generation and workload management

jakubzytka:
This commit introduces comprehensive support for different GPU hardware types in the LLM prompt generation and workload management system:

Key Changes:
- New Generation Profile System: Added PromptGenerationProfile enum and GenerationProfileSpec dataclass to define hardware-specific generation parameters (e.g. A6000 vs A100)
- Database Schema Updates: Added generation_profile field to PromptSeries and executor_class field to SolveWorkload models with corresponding migration
- Multi-Hardware LLM Support: Refactored LLM job generation to support multiple executor classes instead of single hardcoded LLM_EXECUTOR_CLASS
- Profile-Based Prompt Generation: Modified llm_prompt_generation() to iterate over all generation profiles and handle workloads per executor class
- Enhanced Batch Processing: Updated synthetic job processing to support multiple LLM_EXECUTOR_CLASSES
- Test Coverage: Added extensive tests for multi-hardware prompt generation, sampling, and job generation workflows


validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1711,19 +1712,43 @@ def fetch_dynamic_config() -> None:
...
+        if gen_profile == profile
+    ]
+
+    if not executor_classes:
+        logger.info("No executor class uses profile %s; skipping generation", profile)
+        return
+
+    unprocessed_workloads_count = SolveWorkload.objects.filter(
+        finished_at__isnull=True,
+        executor_class__in=executor_classes
```
mzukowski-reef, line 1733: this hack is to allow *all* workloads to finish first - problem is that single workload is like `num_prompts_per_batch` prompt generation runs - so usually it is small number that will run fast - eg. in 5min it can run ~5000 samples which are enough for 5000 synthetic jobs - and generating 5000 batches takes couple of hours - so if we allow them to compete for resources SolveWorkload may be starved for couple of hours - so we simply first allow all workloads to finish and then generate missing prompt series

validator/app/src/compute_horde_validator/validator/tasks.py
```diff
@@ -1711,19 +1712,43 @@ def fetch_dynamic_config() -> None:
...
+
     if unprocessed_workloads_count > 0:
         # prevent any starvation issues
-        logger.info("Unprocessed workloads found - skipping prompt generation")
+        logger.info("Unprocessed workloads found for profile %s - skipping prompt generation", profile)
         return
 
-    num_expected_prompt_series = config.DYNAMIC_MAX_PROMPT_SERIES
-    num_prompt_series = PromptSeries.objects.count()
+    num_expected_prompt_series = config.DYNAMIC_MAX_PROMPT_SERIES_PER_EXECUTOR_CLASS * len(executor_classes)
```
mzukowski-reef, line 1741: this is too much - we need to generate prompt series per generation profile - not per class - classes can reuse generated prompt series - if we need separate prompt series pools of the same size we should create two generation profiles

mzukowski-reef (changes requested):

PR #562, opened 2025-06-18
Add periodic organic task subnet healthcheck (health probe)

kacper-wolkiewicz-reef:


health_probe/health_probe/probe.py
```diff
@@ -0,0 +1,106 @@
...
+        try:
+            return await self.config.client.create_job(job_spec)
+        except ComputeHordeNotFoundError as exc:
+            raise ProbeImproperlyConfigured(
+                "Resource not found while trying to create job. Did you set a proper facilitator url?"
+            ) from exc
+        except ComputeHordeError as exc:
+            raise HealthProbeFailedJobCreation() from exc
+        except Exception as exc:
+            raise HealthProbeFailedJobCreation("Probe encountered unknown exception during job creation.") from exc
```
mpnowacki-reef, line 78: I think this should be propagated to sentry (on top of reporting prometheus metrics). other errors are basically compute horde "supported" like timeouts, rejections, etc. but this can be a bug in the testing code and should warrant a sentry event. I think maybe non-compute horde exception should not be caught here, only in `estimate`, logged together with stack traces with of "error" so sentry issues a slack notification. my idea is that compute horde exceptions issue warnings with stack traces and no sentry events while all other exceptions issue errors with stack traces and sentry events. makes sense? and the same would apply to `_await_job`

PR #569, opened 2025-06-26
CH SDK use of backports for compatibility with Python 3.10

chrisu-inigra:
Conditional imports to support 3.10:
- timezone.utc for UTC
- backports.strenum for StrEnum
- typing_extensions for Self

### DIFF
--- a/compute_horde/pyproject.toml
+++ b/compute_horde/pyproject.toml
@@ -14,7 +14,7 @@ dependencies = [
     "Django~=4.2.4",
     "cryptography>=42.0.8",
     "celery<6,>=5.3.0",
-    "aiohttp>=3.10.10",
+    "aiohttp>=3.12.0",
     "compute-horde-sdk",
     "rich~=13.9.4",
 ]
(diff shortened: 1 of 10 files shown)


compute_horde_sdk/src/compute_horde_core/executor_class.py
```diff
@@ -1,4 +1,9 @@
-from enum import StrEnum
+import sys
+
+if sys.version_info >= (3, 11):  # noqa: UP036
+    from enum import StrEnum
+else:
+    from backports.strenum import StrEnum  # noqa: UP035
```
piotr-figwer-reef, line 6: No strong opinions here, but maybe you can follow `easier to ask for forgiveness than permission` rule and use try/except block here (and in other places that are using `sys.version_info`)?
```suggestion
try:
    from enum import StrEnum
except ImportError:
    from backports.strenum import StrEnum
```

piotr-figwer-reef (approved):
Just one minor nit, other than that looks good!

PR #586, opened 2025-07-15
Fix manifest polling

piotr-figwer-reef:
If miner does not respond with the manifest, we want to copy only the latest manifest, not all of them. This PR fixes the issue with the fallback query, and adds a test to make sure we will not regress in the future.

### DIFF
--- a/validator/app/src/compute_horde_validator/validator/tasks.py
+++ b/validator/app/src/compute_horde_validator/validator/tasks.py
@@ -2147,7 +2147,9 @@ async def _poll_miner_manifests() -> None:
                 m
                 async for m in MinerManifest.objects.filter(
                     miner=miner,
-                ).order_by("-created_at")
+                )
+                .distinct("executor_class")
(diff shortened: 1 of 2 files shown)


kkowalski-reef (commented):
Is there any strong reason for creating available=0 manifests for missing miners? I think it would be easier to just skip miners who don't respond - after all we don't know the actual manifest. AFAIK when selecting manifests for routing we already limit the timeframe so we don't consider old manifests.

piotr-figwer-reef:
> Is there any strong reason for creating available=0 manifests for missing miners? I think it would be easier to just skip miners who don't respond - after all we don't know the actual manifest. AFAIK when selecting manifests for routing we already limit the timeframe so we don't consider old manifests.

We are taking the latest manifest from miner when calculating allowance. So if we skip, we may ending up calculating allowance based on outdated information, and the miner may be e.g. down.

kkowalski-reef (approved):

PR #589, opened 2025-07-16
Fix executor tests

piotr-figwer-reef:
Tests were failing, because:

- `test_artifacts` is producing huge output that GitHub action is not able to handle, causing timeouts.
- GPU mode is not available in CI.

### DIFF
--- a/executor/app/src/compute_horde_executor/executor/tests/integration/test_main_loop.py
+++ b/executor/app/src/compute_horde_executor/executor/tests/integration/test_main_loop.py
@@ -1742,7 +1742,7 @@ def test_artifacts(caplog):
     # Temporarily change logging level. This test produces huge amount of outputs,
     # which GitHub Actions runner is not able to handle, resulting in a timeout.
     caplog.set_level(logging.CRITICAL)
-    
+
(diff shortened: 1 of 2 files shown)


executor/app/src/compute_horde_executor/executor/tests/integration/test_main_loop.py
```diff
@@ -1739,121 +1739,129 @@ def test_multi_volume(httpx_mock: HTTPXMock, tmp_path):
...
-                            "docker_image": "backenddevelopersltd/compute-horde-job-echo:v0-latest",
-                            "timeout_seconds": 10,
-                            "volume_type": "inline",
-                            "job_uuid": job_uuid,
-                            "job_started_receipt_payload": {
-                                "job_uuid": job_uuid,
-                                "miner_hotkey": "miner_hotkey",
-                                "validator_hotkey": "validator_hotkey",
-                                "timestamp": "2025-01-01T00:00:00+00:00",
+    try:
```
kkowalski-reef, line 1742: Found this: https://docs.pytest.org/en/stable/how-to/logging.html#caplog-fixture

kkowalski-reef (approved):
:+1:

PR #590, opened 2025-07-16
error propagation: initial protocol changes

kkowalski-reef:
**This deals with errors in the job flow.**

Key takeout: **errors are strictly categorized as one of: job rejection, job failure or horde failure.**

- New `V0HordeFailedRequest` protocol message communicates ComputeHorde-related issues, `V0JobFailedRequest` will only deal with consumer-actionable failures (job code, volumes, some timeouts)
- SDK will receive the whole status history of a job so that it doesn't miss any status update
- Job status updates contain optional payloads, e.g.: "job failed" will contain the failure stage, reason and stdout/err/return code if available
- Some new try...except safety nets were added and some were tweaked to handle generic error cases (`except Exception: ...`) 
- A lot of common enums were moved to `compute_horde` -> `protocol_consts`
- Lots of random changes to have the new protocol flying and so that random failures are caught and at least propagated as some generic error, most notably:
	- executor JobDriver/JobRunner
	- validator job driving, miner client, facilitator connector, job status updates

### DIFF
--- a/local_stack/send_hello_world_job.py
+++ b/local_stack/send_hello_world_job.py
@@ -186,7 +186,7 @@ async def main() -> None:
 
     # Validate job completion and output.
     expected_artifacts = {'/artifacts/stuff': b'Hello, World!\n'}
-    if job.status != "Completed" or job.result.artifacts != expected_artifacts:
+    if job.status != "completed" or job.result.artifacts != expected_artifacts:
         raise RuntimeError(f"Job failed: status={job.status}, artifacts={job.result.artifacts}")
 
     if verify_http_output_volumes:


executor/app/src/compute_horde_executor/executor/job_driver.py
```diff
@@ -32,81 +48,96 @@ def __init__(self, runner: JobRunner, miner_client: MinerClient, startup_time_li
...
-        try:
-            await self._do_execute()
-
-        except TimeoutError:
-            logger.warning(f"Job timed out during {self.current_stage} stage")
-            await self.miner_client.send_job_error(
-                JobError(
-                    f"Timed out during {self.current_stage} stage",
-                    error_type=V0JobFailedRequest.ErrorType.TIMEOUT,
+        async with self.miner_client:  # TODO: Can this hang?
```
mpnowacki-reef, line 58: it's network, of course it can

executor/app/src/compute_horde_executor/executor/job_driver.py
```diff
@@ -32,81 +48,96 @@ def __init__(self, runner: JobRunner, miner_client: MinerClient, startup_time_li
...
-                self._set_deadline(timing_details.allowed_leeway, "allowed leeway")
-            elif initial_job_request.timeout_seconds is not None:
-                # For single-timeout, this is the full timeout for the whole job
-                self._set_deadline(initial_job_request.timeout_seconds, "single-timeout mode")
-            else:
-                raise JobError(
-                    "No timing received: either timeout_seconds or timing_details must be set"
-                )
+        # Download stage
+        if timing_details:
```
mpnowacki-reef, line 115: we should probably add a todo to remove `if timing_details:`, right?

kkowalski-reef: In the protocol, the fine grained timing details are still optional. Synthetic jobs use the single timeout value. And in the back of my head I have the possibility of allowing SDK users to just specify a single timeout if they don't care enough to do the profiling - may be important to reduce the fricton of hordizing.

But the timing code could be reworked so that it does not pollute the job flow that much.

PR #599, opened 2025-07-19
Add --job-runner-class argument to run_executor manage.py command.

kacper-wolkiewicz-reef:


miner/app/src/compute_horde_miner/settings.py
```diff
@@ -493,3 +494,16 @@ def BITTENSOR_WALLET() -> bittensor.wallet:
 
 CLUSTER_SECRET = env.str("CLUSTER_SECRET", default="")
 CLUSTER_HOTKEYS = env.list("CLUSTER_HOTKEYS", default=[])
+
+# Path pointing to the file containing custom job runner classes.
+CUSTOM_JOB_RUNNERS_PATH = ""
```
mpnowacki-reef, line 499: this is essentially only for the non-production executor-managers, right? like dockerexecutormanager and devexecutormanager? I'm thinking leaving these settings here might be misleading, maybe the env vars should be read directly in the modules with those managers?

executor/app/src/compute_horde_executor/executor/job_runner.py
```diff
@@ -245,113 +237,26 @@ async def start_job(self) -> AsyncGenerator[None, Any]:
...
-                        self.nginx_container_name,
-                    )
-                    try:
-                        await asyncio.wait_for(process.wait(), DOCKER_STOP_TIMEOUT_SECONDS)
-                    except TimeoutError:
-                        process.kill()
-                        raise
-                except Exception as e:
-                    logger.error(f"Failed to stop Nginx: {e}")
+            yield await self.after_start_job()
```
mpnowacki-reef, line 258: shouldn't this be 
```suggestion
            await self.after_start_job()
            yield 
```
?
I'm not sure the value returned by `self.after_start_job` should be passed up as the value of this async context manager. previously it was none.

mpnowacki-reef (changes requested):
Aside from the two minor comments I've left the changes are nice and concise and generally good enough for merging. The documentation part however is lacking. It should be stated clearly how miners can tweak their executors. They need to rely on our code to do that (inherit from some class) so it should be stated clearly what is part of the public interface. I also think the default job runner class should have some docstrings explaining that it facilitates streaming (we may skip the raw script here as we'll be removing it) and otherwise copies the consumer-provided cmd and docker image.

miner/app/src/compute_horde_miner/settings.py
kacper-wolkiewicz-reef: I moved the docker/dev executor related settings to stick together.

PR #600, opened 2025-07-21
Volume manager support

mateus-cardoso-reef:


local_stack/send_huggingface_job.py
```diff
@@ -0,0 +1,74 @@
+import asyncio
```
mpnowacki-reef, line 1: this script isn't used anywhere in tests or whatever, right? I think it's gonna rot in such case.

compute_horde_sdk/src/compute_horde_core/volume/_manager.py
```diff
@@ -0,0 +1,172 @@
...
+            Parsed JSON response
+
+        Raises:
+            VolumeManagerError: If the request fails or response is invalid
+
+        """
+        logger.debug(f"Making {operation} request to Volume Manager at {url}")
+
+        try:
+            async with httpx.AsyncClient(timeout=timeout) as client:
```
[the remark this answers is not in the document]
mateus-cardoso-reef: Ok, implementing that

executor/app/src/compute_horde_executor/executor/job_runner.py
```diff
@@ -242,6 +258,17 @@ async def start_job(self) -> AsyncGenerator[None, Any]:
                 f"{self.artifacts_mount_dir.as_posix()}/:{self.full_job_request.artifacts_dir}",
             ]
 
+        # Build volume mount flags
+        volume_flags = []
+        if self.volume_manager_mounts:
+            # Use volume manager mounts
+            for mount in self.volume_manager_mounts:
+                volume_flags.extend(["-v", f"{mount.source}:{mount.target}"])
```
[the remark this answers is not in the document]
mateus-cardoso-reef: You're right, it doesn’t. I implemented support for bind mounts with only the `type=bind` option, since that’s what was already used in the direct download path (which doesn’t go through the Volume Manager). But your comment gave me the idea to slightly modify the Volume Manager integration so that it only provides the mount options as a list of strings, which covers all possible choices. This makes the changes in `ComputeHorde` much smaller and cleaner and leaves the task of generating the appropriate docker mount options to the Volume Manager.

local_stack/prepare.sh
```diff
@@ -61,6 +61,8 @@ PROJECT_ROOT=$(git rev-parse --show-toplevel)
 cd $PROJECT_ROOT/executor
 ./setup-dev.sh
 update_env_var "DEBUG_NO_GPU_MODE" "true" ".env"
+# Use for volume manager testing, not a requirement for the local stack setup
+update_env_var "VOLUME_MANAGER_ADDRESS" "http://localhost:8001" ".env"
```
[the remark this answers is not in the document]
mateus-cardoso-reef: Actually, nothing would break, since the `VolumeManagerClient` gracefully handles cases where the Volume Manager doesn’t exist, so that’s not a problem. That said, I noticed the tests were a bit slower with that configuration and also produced unnecessary warning logs. So I’m commenting out that line and adding a note in the `README.md` with instructions on how to enable it if needed.

local_stack/send_huggingface_job.py
mateus-cardoso-reef: I originally implemented this script thinking it might be useful for anyone testing volume downloads, but it turns out it wasn’t mentioned in any `README.md` and had a fairly narrow use case. So I’m removing the script and replacing it with an argparse `--test-volume` flag. I’ve also updated the `README.md` with instructions on how to use it.

compute_horde_sdk/src/compute_horde_core/volume/_manager.py
```diff
@@ -83,9 +80,14 @@ async def prepare_volume(
         volume_data = volume.model_dump()
         payload = {"job_uuid": job_uuid, "volume": volume_data, "job_metadata": job_metadata}
 
-        response_data = await self._make_request(url, payload, "prepare_volume")
-        mounts = [VolumeManagerMount(**mount) for mount in response_data["mounts"]]
-        return VolumeManagerResponse(mounts=mounts)
+        response_data = await self._make_request(url, payload, "prepare_volume", timeout=timeout)
+
+        # Convert string mount types to MountType objects
```
mpnowacki-reef, line 85: I don't see any such conversion here, what am I missing?

compute_horde_sdk/src/compute_horde_core/volume/_manager.py
```diff
@@ -83,9 +80,14 @@ async def prepare_volume(
...
+        mounts = []
+        for mount_data in response_data["mounts"]:
+            mounts.append(mount_data)
+
+        return mounts
```
mpnowacki-reef, line 90:
```suggestion
        return response_data["mounts"]
```

compute_horde_sdk/tests/unit/core/test_volume.py
```diff
@@ -555,9 +554,11 @@ async def test_prepare_volume_success(self, client, huggingface_volume):
                 job_uuid="test-job-123", volume=huggingface_volume, job_metadata={"image": "test-image"}
             )
 
-            assert isinstance(result, VolumeManagerResponse)
-            assert len(result.mounts) == 1
-            assert result.mounts[0].type == "bind"
+            assert isinstance(result, list)
```
mpnowacki-reef, line 557: is this AI? this looks like an unsupervised agent left for testing. why isn't this just 
```
assert result == [["-v", "/host/path:/container/path"]]
```

compute_horde_sdk/tests/unit/core/test_volume.py
```diff
@@ -642,23 +645,19 @@ async def test_prepare_volume_with_multiple_mounts(self, client, huggingface_vol
...
+            assert all(isinstance(item, list) for item in result)
+            assert all(all(isinstance(subitem, str) for subitem in item) for item in result)
+            assert len(result) == 3
 
             # Check first mount
-            assert result.mounts[0].type == "bind"
-            assert result.mounts[0].source == "/host/models"
-            assert result.mounts[0].target == "/volume/models"
+            assert result[0] == ["-v", "/host/models:/volume/models"]
```
mpnowacki-reef, line 654: what's happening here?

local_stack/send_hello_world_job.py
```diff
@@ -81,54 +88,104 @@ def create_mt_client_with_tempfiles(
...
 
 
-async def main() -> None:
-    # Pull required Docker images
-    os.system("docker pull alpine")
-    os.system("docker pull backenddevelopersltd/compute-horde-executor:v1-latest")
-    os.system("docker pull backenddevelopersltd/compute-horde-streaming-job-test:v0-latest")
-
-    compute_horde_streaming_job_spec = ComputeHordeJobSpec(
+async def run_volume_test():
```
mpnowacki-reef, line 95: I think you should include the POC volume manager in this repo (in the right place, a path explaining that this is a POC), start in this test and then have the test run in CI. End to End

mpnowacki-reef (changes requested):

executor/app/src/compute_horde_executor/executor/job_runner.py
```diff
@@ -448,32 +496,68 @@ async def _unpack_volume(self, volume: Volume | None):
...
+        assert self.volume_manager_client is not None
+
+        job_uuid = self.initial_job_request.job_uuid
+
+        # Prepare job metadata
+        job_metadata = self.full_job_request.model_dump()
+
+        try:
+            logger.debug(f"Requesting volume preparation from Volume Manager for job {job_uuid}")
+            response = await self.volume_manager_client.prepare_volume(
```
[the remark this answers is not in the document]
kkowalski-reef: if you mean the total timeout to download the volumes for the job - as specified by the consumer in job spec - this is achieved with overarching timeout for the whole download stage in the job driver. if that timeout triggers, the task will be cancelled. we don't pass the timeout value around. if the cancellation is somehow swallowed, the validator will give up waiting for the download anyway and there is (or was supposed to be) a "reaper" for the executor process in case it hangs because of this.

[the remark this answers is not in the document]
kkowalski-reef: this is coming from their allowance so if they want to and can afford it...

compute_horde_sdk/src/compute_horde_core/volume/_manager.py
mateus-cardoso-reef: You're right. This originally came from the previous version where I was constructing `VolumeManagerResponse` with `VolumeManagerMount` objects, the loop now is unnecessary.
Cleaning it up.

executor/app/src/compute_horde_executor/executor/job_runner.py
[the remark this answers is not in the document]
kkowalski-reef: Would be useful to also mention this in the docs - the manager service has to not time out itself during the preparation of the volumes.

compute_horde_sdk/README.md
```diff
@@ -247,6 +247,109 @@ The SDK includes an optional `FallbackClient` that mirrors the standard `Compute
...
+export VOLUME_MANAGER_ADDRESS="http://localhost:8080"
+```
+
+You can also add custom headers for authentication or any other purpose. Any environment variable starting with `VOLUME_MANAGER_HEADER_` will be added as a header to each request sent to the volume manager.
+
+**Examples:**
+
+```bash
+# Authentication header
+export VOLUME_MANAGER_HEADER_AUTHORIZATION='Bearer dupadupakupa'
```
mpnowacki-reef, line 277:
```suggestion
export VOLUME_MANAGER_HEADER_AUTHORIZATION='Bearer tokentokentoken'
```

executor/app/src/compute_horde_executor/executor/job_runner.py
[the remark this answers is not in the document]
mateus-cardoso-reef: Okay, so I will be removing the timeout from volume manager client, leaving just the `connect=30.0`  for safety. I'm also adding a **note** in the docs about the issue.

compute_horde_sdk/README.md
```diff
@@ -247,6 +247,101 @@ The SDK includes an optional `FallbackClient` that mirrors the standard `Compute
...
+Your volume manager must implement these endpoints:
+
+#### `POST /prepare_volume`
+
+**Request:**
+```json
+{
+  "job_uuid": "7b522daa-e807-4094-8d96-99b9a863f960",
+  "volume": {
+    "volume_type": "huggingface_volume",
```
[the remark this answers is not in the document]
mateus-cardoso-reef: Just a quick update, I double-checked it and it actually always sends the volume wrapped in MultiVolume, that happens in the `models.py` from the facilitator

PR #604, opened 2025-07-30
Fix fallback client instabilities

piotr-figwer-reef:


compute_horde_sdk/pyproject.toml
```diff
@@ -40,6 +40,7 @@ fallback = [
     "wheel",
     "setuptools",
     "skypilot[runpod]>=0.8.1,<0.9.0",
+    "gpustat<1.1",
```
piotr-figwer-reef, line 43: Tests were failing with ``× Failed to build `gpustat==1.1.1``.

compute_horde_sdk/src/compute_horde_sdk/_internal/fallback/job.py
```diff
@@ -173,6 +173,15 @@ async def wait_for_streaming(self, timeout: int = 120) -> None:
         :raises FallbackJobTimeoutError: If streaming does not start within timeout seconds.
         """
         start_time = time.time()
+
+        while self.status != FallbackJobStatus.ACCEPTED:
```
piotr-figwer-reef, line 177: Streaming server can start up during setup phase, before the job is actually ready. Let's wait for job to be `ACCEPTED` before we return the information that streaming server is ready.

mpnowacki-reef (approved):

PR #615, opened 2025-08-07
Replace shell commands with Docker bindings

jan-sauer-reef:
Instead of crafting shell commands to interact with Docker, the code now uses Docker bindings (aiodocker for full async support)

COM-547 Use Docker bindings for Python instead of crafting bash commands


executor/app/src/compute_horde_executor/executor/utils.py
```diff
@@ -105,33 +137,66 @@ def get_machine_specs() -> MachineSpecs:
...
-        program: Program to execute
-        *args: Program arguments
+        image: Docker image to run
+        name: Name of the container (default: None)
+        command: Command to execute in the container. This should be formatted as a list that the
+            Docker API would understand, e.g. ["bash", "-c", "..."]. If None, will run the default
+            command for the image (default: None)
         clean_exit_timeout: Seconds to wait before force kill (default: 1.0)
-        **subprocess_kwargs: Additional keyword arguments passed to asyncio.create_subprocess_exec()
+        auto_remove: Automatically remove the container on exit. Equivalent to --rm flag (default: False)
```
piotr-figwer-reef, line 159: It's worth adding `"HostConfig": {"AutoRemove": true}`.

jan-sauer-reef: Race conditions appear to cause problems with reading logs from Docker containers when setting AutoRemove to true with aiodocker. I wasn't able to figure out how to reliably grab them, so I think using the context manager to handle the container lifecycle might be better.

piotr-figwer-reef (approved):
LGTM!

PR #627, opened 2025-08-14
Receipts module

piotr-figwer-reef:


validator/app/src/compute_horde_validator/validator/management/commands/transfer_receipts.py
```diff
@@ -97,217 +42,10 @@ async def handle(
...
-
-    async def _throw_if_disabled(self):
-        try:
-            if await aget_config("DYNAMIC_RECEIPT_TRANSFER_ENABLED"):
-                return
-        except KeyError:
-            logger.warning("DYNAMIC_RECEIPT_TRANSFER_ENABLED dynamic config is not set up!")
-
-        raise TransferIsDisabled
+        await Receipts().run_receipts_transfer(
```
kkowalski-reef, line 45: afaik the pattern is to use the singleton thingy like in allowance? (`def receipts() -> Receipts: ...`)

kkowalski-reef (changes requested):
@piotr-figwer-reef I'm seeing a ton of random weird things in the code:

- sync_to_async on django querysets when we django has awaitable methods for queries
- I'm not following why some methods are wrapped in a try...except and return `None` / `[]` instead of throwing
- the error handling is in general inconsistent, `create_job_started_receipt` wraps exceptions into `ReceiptsGenerationError`, `create_job_finished_receipt` doesn't
- lots of verbose debug logging - do we want to know about every time receipts are retrieved? even for DEBUG this seems too much
- `raise ValueError(f"Expected datetime, got {type(ts)}")` is unreachable, `supertensor().get_block_timestamp(block_number)` always returns a datetime according to the typing, if not that should be fixed instead

Please look through the code and address these issues first.

validator/app/src/compute_horde_validator/validator/management/commands/transfer_receipts.py
piotr-figwer-reef: Not sure if the singleton thingy is useful here, but added it for consistency.

PR #632, opened 2025-08-18
Pull up common error types to compute_horde

kkowalski-reef:
- Pulled out the 3 main error classes to compute_horde. This way we can throw these during job execution (almost?) anywhere and they should be handled correctly.
- (still, throwing an error where it is not appropriate, e.g. a rejection in an executor JobDriver, will be treated as a generic horde failure)
- Added a helper for wrapping unhandled exceptions into a HordeError, so that they are a bit more consistent
- Explicitly logging exceptions to sentry where they would've been only wrapped and propagated. They are potentially interesting to log.

### DIFF
--- a/validator/app/src/compute_horde_validator/validator/organic_jobs/facilitator_client.py
+++ b/validator/app/src/compute_horde_validator/validator/organic_jobs/facilitator_client.py
@@ -7,6 +7,7 @@ from typing import Any, Literal
 import bittensor_wallet
 import httpx
 import pydantic
+import sentry_sdk
 import tenacity
 import websockets
 from channels.layers import get_channel_layer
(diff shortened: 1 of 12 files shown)


validator/app/src/compute_horde_validator/validator/organic_jobs/facilitator_client.py
```diff
@@ -302,9 +304,10 @@ async def handle_job_status_updates(self, job_uuid: str):
                         return
                 except pydantic.ValidationError as exc:
                     logger.warning("Received malformed job status update: %s", exc)
-        except Exception:
+        except Exception as e:
+            # Nothing that gets thrown here is expected.
+            sentry_sdk.capture_exception(e)
             logger.exception("Error in job status update listener", exc_info=True)
```
emnoor-reef, line 310: Won't sentry catch this twice? Since logger.exception is basically an ERROR log with stack trace. Maybe we can lower the level to WARNING?

kkowalski-reef: yep, looks like it

PR #655, opened 2025-08-21
test wallet module

adalc-reef:


compute_horde/compute_horde/test_wallet.py
```diff
@@ -0,0 +1,80 @@
...
+
+MINER_WALLET_NAME = f"test_miner{_UNIQUE_WALLET_TAG}"
+MINER_WALLET_HOTKEY = "default"
+
+VALIDATOR_WALLET_NAME = f"test_validator{_UNIQUE_WALLET_TAG}"
+VALIDATOR_WALLET_HOTKEY = "default"
+
+
+@cache
+def get_misc_wallet() -> bittensor_wallet.Wallet:
```
kkowalski-reef, line 24: Can we call these functions something like "get_test_misc_wallet()"? They will show up in autocomplete

kkowalski-reef (approved):
There are some leftovers that just pass through a wallet like these:
```python
@pytest.fixture(scope="session", autouse=True)
def wallet():
    return get_validator_wallet()
```

Let's get rid of them?

PR #700, opened 2025-09-10
New agent guidelines

adalc-reef:


kkowalski-reef:
let's keep the guildelines in somethign like `GUIDELINES.md` in the root of the repo and point everything else to that.

.junie/guidelines.md
```diff
@@ -18,6 +28,8 @@ This project has several subprojects:
 5. compute_horde_sdk
 6. compute_horde lib
 
+This project uses Astral uv. Use `uv run` to run Python code or Python tools.
```
kkowalski-reef, line 31:
```suggestion
This project uses Astral uv. Use `uv run` to run Python code or Python tools. Do not try to create or activate a venv manually.
```

PR #708, opened 2025-09-16
test(validator): COM-713 reduce routing test mocking

adalc-reef:


validator/app/src/compute_horde_validator/validator/routing/tests/test_job_routing.py
```diff
@@ -144,45 +147,51 @@ async def test_pick_miner_for_two_long_jobs(add_allowance):
 
 @pytest.mark.django_db(transaction=True)
 @pytest.mark.asyncio
-async def test_pick_miner_for_job__skips_busy_miner_based_on_receipts(add_allowance, monkeypatch):
-    # Pick the first suitable miner and mark it as busy with as many ongoing jobs
-    # as its executor_count for the requested class so it gets skipped.
+async def test_pick_miner_for_job__skips_busy_miner_based_on_receipts(add_allowance):
+    """If a miner is saturated by started receipts, routing skips it."""
+    # Clean any leftover receipts
+    await JobStartedReceipt.objects.all().adelete()
```
piotr-figwer-reef, line 153: I can see that you’re putting those cleanup steps at the beginning of each test. Are they actually needed because other tests are leaving the database in an unclean state? Or are they just there to be extra cautious?

piotr-figwer-reef (approved):

PR #723, opened 2025-09-29
scoring metrics

kkowalski-reef:


kkowalski-reef:
not necessary; already cherry-picked where needed

PR #835, opened 2026-01-28
feat: make organic job execution flow sync

emnoor-reef:


validator/app/src/compute_horde_validator/validator/job_excuses.py
```diff
@@ -114,3 +114,30 @@ async def get_expected_miner_executor_count(
...
+    miner_hotkey: str,
+    executor_class: ExecutorClass,
+) -> int:
+    latest_manifest = (
+        MinerManifest.objects.filter(
+            miner__hotkey=miner_hotkey,
+            executor_class=executor_class,
+            created_at__lte=check_time,
+        )
+        .order_by("created_at")
```
mpnowacki-reef: I think this is mistake that the ordering is not `-created_at`, here and in the async version.

validator/app/src/compute_horde_validator/validator/organic_jobs/miner_driver_sync.py
```diff
@@ -0,0 +1,1149 @@
...
+                msg.__class__.__name__,
+                expected_message_type.__name__,
+                self.job.job_uuid,
+            )
+
+    def run(self) -> None:
+        try:
+            self._run()
+        except Exception as exc:
+            sentry_sdk.capture_exception(exc)
```
mpnowacki-reef, line 624: why is the sentry_sdk called explicitly here, instead of just logging the exception, possibly with job info?

validator/app/src/compute_horde_validator/validator/tests/test_organic_jobs/test_miner_driver_sync.py
```diff
@@ -0,0 +1,339 @@
...
+        volume=v2_request.volume,
+        output_upload=v2_request.output_upload,
+        download_time_limit=v2_request.download_time_limit,
+        execution_time_limit=v2_request.execution_time_limit,
+        streaming_start_time_limit=v2_request.streaming_start_time_limit,
+        upload_time_limit=v2_request.upload_time_limit,
+    )
+
+
+@pytest.mark.django_db(databases=["default", "default_alias"], transaction=True)
```
mpnowacki-reef, line 104: I really like how concise these tests are. readable and really doing the job.

validator/app/src/compute_horde_validator/validator/organic_jobs/miner_driver_sync.py
```diff
@@ -0,0 +1,1149 @@
...
+        )
+
+        self._record_event(
+            type=SystemEvent.EventType.MINER_ORGANIC_JOB_FAILURE,
+            subtype=subtype,
+            long_description=comment,
+        )
+        return DriverState.FAILED
+
+    def _wait_for_message(
```
mpnowacki-reef, line 517: seems that this implementation is very robust and resistant against weird messages, out of order messages etc. but I think it is not currently tested against that, is it? We've previously had malicious miners gaining some edge by sending out of order messages.

mpnowacki-reef (changes requested):
The implementation looks functionally correct, but SyncOrganicJobDriver at ~800 lines is doing a lot of heavy lifting. What do you think about splitting it? A few ideas:

Message handling - The logic in _wait_for_message that handles different error types (lines 477-537) could live in MinerClient or a separate helper. The client already knows how to send/receive, it could also know how to classify responses.

Receipt operations - Receipt creation and sending is scattered across _reserve_executor, _send_job_accepted_receipt, _collect_results, and _send_job_finished_receipt_if_failed. A small ReceiptService or even just a few standalone functions could clean this up.

Failure handling - _handle_horde_failure, _handle_decline_job, _handle_job_failed, _handle_horde_failed (~170 lines) are cohesive and could be extracted.

This would leave the driver as a ~300-400 line state machine focused on orchestration.
Also, since this is becoming a critical path, could we add some docstrings to the main flow? Specifically:

A class-level docstring explaining the state machine and the job lifecycle stages
Brief docs on _run() and _wait_for_message() explaining how the state transitions work

The state enum is already self-documenting which is great, just want to make sure the next person can follow the flow quickly.

validator/app/src/compute_horde_validator/validator/job_excuses.py
emnoor-reef: yes, I noticed that, but kept it like that for the time being to align with the async impl in the first iteration.

validator/app/src/compute_horde_validator/validator/organic_jobs/miner_driver_sync.py
emnoor-reef: That I have copied verbatim from the async impl :v

validator/app/src/compute_horde_validator/validator/organic_jobs/miner_driver_sync.py
```diff
@@ -0,0 +1,1180 @@
...
+        comment: str,
+        event_subtype: SystemEvent.EventSubType,
+        horde_failure_reason: HordeFailureReason,
+        reported_by: JobParticipantType = JobParticipantType.VALIDATOR,
+        horde_failure_context: FailureContext | None = None,
+        incident_type: MinerIncidentType | None = None,
+        log_exc_info: bool = False,
+    ) -> DriverState:
+        logger.warning(comment, exc_info=log_exc_info)
+        self.job.status = OrganicJob.Status.FAILED
```
kkowalski-reef, line 354:
```suggestion
        self.job.status = OrganicJob.Status.HORDE_FAILED
```

It worries me that no test caught this

validator/app/src/compute_horde_validator/validator/organic_jobs/miner_driver_sync.py
```diff
@@ -0,0 +1,1180 @@
...
+                # NOTE: No incident reported - timeouts are not miner incidents,
+                # consistent with the async implementation.
+                return self._handle_horde_failure(
+                    comment=f"Timed out waiting for miner (reason={timeout_status_reason})",
+                    event_subtype=SystemEvent.EventSubType.ERROR_VALIDATOR_REPORTED_TIMEOUT,
+                    horde_failure_reason=timeout_status_reason,
+                )
+
+            try:
+                msg = self.miner_client.recv()
```
kkowalski-reef: Can we pass in the remaining timeout to recv? There is this timer utility that you can set up before the loop and do a ~ `recv(timeout=deadline.time_left())`. Otherwise we're only checking for the timeout every time recv returns with no response which seems to be every 5 seconds. The job timing is strict and I'm not even sure how this will influence it but I can imagine we may end up with one stage consuming the time of a subsequent stage or something like that, not fun to debug.

validator/app/src/compute_horde_validator/validator/organic_jobs/miner_driver_sync.py
```diff
@@ -0,0 +1,1180 @@
...
+                    reason=HordeFailureReason.STREAMING_FAILED,
+                    message="Executor reported legacy V0StreamingJobNotReadyRequest message",
+                )
+            elif isinstance(msg, V0ExecutorFailedRequest):
+                msg = V0HordeFailedRequest(
+                    job_uuid=msg.job_uuid,
+                    reported_by=JobParticipantType.MINER,
+                    reason=HordeFailureReason.GENERIC_ERROR,
+                    message="Executor reported legacy V0ExecutorFailedRequest message",
+                )
```
kkowalski-reef: These are not being sent any more. This was only needed for the release of the error reporting. You can safely remove this.

validator/app/src/compute_horde_validator/validator/organic_jobs/miner_driver_sync.py
```diff
@@ -0,0 +1,1180 @@
...
+                    self._state = self._collect_results()
+                case DriverState.COMPLETE:
+                    return
+                case DriverState.FAILED:
+                    return
+                case _:
+                    assert_never(self._state)
+
+    def _connect(self) -> DriverState:
+        assert self._state == DriverState.CONNECT
```
kkowalski-reef, line 733: Any reason for this defensive pattern throughout the driver?

kkowalski-reef:
I don't think there's much that can be done here to lower the LoC. We are passing around a lot of context into messages, events, error reports etc and extracting anything means that you still have to pass that stuff into the helper, or if you instead give it the whole job driver instance, this is just method calling with extra steps. Some state methods are literally just "send a message" so not much to gain here. Maybe some methods could be split into less of "how to build a receipt object" and more of "the receipt should be sent now". But IMO looking at the code of the state hanlers, on average it's not worth it. I'd rather have code related to a state be in one place, no single method is hard to comprehend as-is.

Still, I'm not a fan of the state handler deciding the next state. I would maybe structure the "next state decision" and error handling a bit differently, e.g. states are for the happy path, otherwise throw exceptions. I looked into it and it seems I already did something like this for the async driver, there are generic exceptions like HordeError and more meaningful exceptions like "MinerRejectedJob", "MinerReportedJobFailed" etc, makes it so that:
- you have a wrapped high level "happy path" list of steps
- a separate list of "except if this happens, that's how to report the problem before exiting cleanly"
- there's no need to bubble up errors, gets rid of the `if isinstance(result, DriverState)` checks
- these exception classes already contain a lot of useful context to report so could make it easier to handle them in a generic way (see the error handling in the async driver)

But this is not critical.

kkowalski-reef (dismissed):

validator/app/src/compute_horde_validator/validator/organic_jobs/miner_driver_sync.py
emnoor-reef: `OrganicJob.Status` doesn't have a `HORDE_FAILED`.

validator/app/src/compute_horde_validator/validator/organic_jobs/miner_driver_sync.py
emnoor-reef: just for peace of mind :D

mpnowacki-reef (approved):

